PR #501, opened 2014-04-29
Jds7717 changes

jds7717:
This will apply the finalScale to the scale from the KML; before it was ignored


jasonbeverage:
Included in 1239c21844cc632b321b0e59e65901da6d37f116

PR #541, opened 2014-08-22
Complete quoting for parameters of some CMake commands.

elfring:
[A wiki article pointed out](http://cmake.org/Wiki/CMake/Language_Syntax#CMake_splits_arguments_unless_you_use_quotation_marks_or_escapes. "CMake splits arguments unless you use quotation marks or escapes.") that whitespace will only be preserved for parameters in CMake commands if passed strings will be appropriately quoted or escaped.

Quoting can be added so that more places should also handle file names correctly which contain semicolons eventually.


elfring:
I would appreciate a bit more constructive feedback for [this update suggestion](https://github.com/gwaldron/osgearth/pull/541/commits/737b427abe368920563c723e40527526785d73bf "More quoting for parameters of some CMake commands").

jasonbeverage:
Hi Markus,

I closed this b/c it was out of date and didn't merge cleanly.  We'll need to do another run through on this to get it merged with the current master.

elfring:
Would you like to update the shown approach for the current build scripts?

jasonbeverage:
Markus that would be great if you don't mind creating another PR, that
would be really helpful.

PR #596, opened 2015-03-04
Map::calculateProfile: fix fallback/default logic

quixotique:
Was not checking all the image and elevation layers for a valid profile before falling back to the default (WGS84 for geocentric, etc.)

I have an application whose TileSource has an unusual geocentric projection, and I need that ImageLayer to determine the Map's spatial reference.  Without this pull request, Map::calculateProfile() always chooses WGS84.  With this pull request, in the absence of a configured Map profile, any image or elevation layer that supplies a profile of the correct kind will determine the Map's profile.

### DIFF
--- a/src/osgEarth/Map.cpp
+++ b/src/osgEarth/Map.cpp
@@ -962,25 +962,41 @@ Map::calculateProfile()
                         << std::endl;
                 }
             }
-
-            if ( !_profile.valid() )
-            {
-                // by default, set a geocentric map to use global-geodetic WGS84.
-                _profile = osgEarth::Registry::instance()->getGlobalGeodeticProfile();
-            }
         }
-
         else if ( _mapOptions.coordSysType() == MapOptions::CSTYPE_GEOCENTRIC_CUBE )
         {
-            //If the map type is a Geocentric Cube, set the profile to the cube profile.
-            _profile = osgEarth::Registry::instance()->getCubeProfile();
+            if ( userProfile.valid() )
+            {
+                if ( userProfile->isOK() && userProfile->getSRS()->isCube() )
+                {
+                    _profile = userProfile.get();
+                }
+                else
+                {
+                    OE_WARN << LC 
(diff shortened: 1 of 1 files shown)


gwaldron:
Can you elaborate on this unusual geocentric projection?
Just want to fully understand the justification here. Thanks.

quixotique:
We have a legacy application with its own scene graph library that projects maps and other data onto a unit sphere instead of an earth-scale-in-metres ellipsoid.  I am trying to replace the map display component with osgEarth whilst retaining all the other display layers of the existing scene graph.  In order to do that I need to set the osgEarth::Map Profile to use a non-standard spherical ECEF geographic SpatialReference defined by a PROJ4 parameter string of our own devising.

I set up a custom TileSource for our map images, and place it into an ImageLayer which I initialise with our custom Profile that uses our non-standard SpatialReference.  However, when I instantiate a Map object using that ImageLayer, the Map always sets up its Profile as WGS84 ellipsoid, and as a result, the osgEarth layer and legacy layers do not coincide.

There is no explicit Map::setProfile() method.  The only way to explicitly set a Map's profile is through its config options, which in our case is cumbersome since we do everything in code and do not want to put osgEarth XML config files into our product.

Reading the osgEarth code, it appeared that the intention was to allow layers to determine a Map's profile before the Map applied its own default/fallback logic.  I have simply implemented that intention more correctly.  If that was never the intention, then my pull request is not useful.

If you look at the code I have written, it does not add any "unusual" spatial references to the code, it just allows their use without losing the existing default choices.  It simply allows image and elevation layers to determine a Map's profile *before* the Map falls back to WGS84.  I also generalised that logic to other kinds of SRS: geocentric cube and projection, so that behaviour is consistent for all three.

gwaldron:
Thanks for the details. I'll merge it in now.
For the record, there is no Map::setProfile() method, but you can set the profile when creating the Map with the MapOptions::profile property.

PR #607, opened 2015-04-29
Corrections needed to compile with GCC 4.8

BrandonTheHamm:
I had to make a few minor changes to enable compilation with GCC 4.8.


gwaldron:
Thanks - I think we have these fixed already.

PR #617, opened 2015-06-22
Topic rex2

pprabhu78:
Morphing Implementation

Use test-morphing.earth in osgearth viewer and use key 1 and then zoom in and out


gwaldron:
Merged manually outside of github. Thanks.

PR #619, opened 2015-06-25
Fixes to use VerticalDatums with different units

AlexBobkov:
Comments:

src/osgEarth/GeoData.cpp
_srs has no VerticalDatum
rhs.getSRS() has VerticalDatum
So only horizontal srs should be compared

src/osgEarth/VerticalDatum
virtual methods for more freedom in VerticalDatum subclassing

src/osgEarth/VerticalDatum.cpp
usually we want output units to be meters

 src/osgEarthDrivers/gdal/ReaderWriterGDAL.cpp
these two deleted lines restrict raw heights and duplicate
if (v < getNoDataMinValue()) return false;
if (v > getNoDataMaxValue()) return false;

### DIFF
--- a/src/osgEarth/GeoData.cpp
+++ b/src/osgEarth/GeoData.cpp
@@ -1089,7 +1089,7 @@ GeoExtent::expandToInclude( const GeoExtent& rhs )
 GeoExtent
 GeoExtent::intersectionSameSRS( const GeoExtent& rhs ) const
 {
-    if ( isInvalid() || rhs.isInvalid() || !_srs->isEquivalentTo( rhs.getSRS() ) )
+    if ( isInvalid() || rhs.isInvalid() || !_srs->isHorizEquivalentTo( rhs.getSRS() ) )
         return GeoExtent::INVALID;
 
     if ( !intersects(rhs) )
--- a/src/osgEarth/VerticalDatum
+++ b/src/osgEarth/VerticalDatum
@@ -82,13 +82,13 @@ namespace osgEarth
          * Converts an MSL value (height relative to a mean sea level model) to the
          * corresponding HAE value (height above the model's reference ellipsoid)
          */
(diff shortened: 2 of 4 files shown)


jasonbeverage:
Thanks for the changes.  I've merged this but then modified it a bit.  GDAL can report the linear units of a spatial reference, so I added a change to the GDAL driver that should always convert values to meters before it's returned.  So we can keep the check for values that are out of range.

AlexBobkov:
Jason, check against noData value from the raster band should be done for the raw height value before scaling.

jasonbeverage:
Good catch.  I've pushed a fix for that, and went ahead and took your change to remove the checks against 32000 and -32000.  You're right, those do duplicate the logic with getNoDataMinValue and getNoDataMaxValue.  I suppose for a millimeter dataset though you'll need to have to manually set those values since the defaults are assuming meters.

AlexBobkov:
Jason, I've tested your change with linear units from OSRGetLinearUnits. It doesn't work because these units are from horizontal srs. GDAL documentation says:

>This method only checks directly under the PROJCS, GEOCCS or LOCAL_CS node for units.

However there is another way: function GDALGetRasterUnitType returns units of the values in the raster band. Units may be "m", "mm", "cm".

jasonbeverage:
Doh.  I made a similar change to one of our other products that uses GDAL
to support datasets in feet and it worked fine.  That was probably b/c the
whole dataset was in feet or something and wasn't the correct fix.  I've
just removed the _linearUnits stuff from the GDAL driver for now.  If
you're good with your overrides for the vertical datum then we'll just
leave it be for now until we can think it out a bit more.

AlexBobkov:
It works now for me with my VerticalDatum with minor fix from another pull request: https://github.com/gwaldron/osgearth/pull/621

PR #623, opened 2015-07-01
Remove radian conversion when using SilverLining SetLongitude and SetLatitude functions

cgpoh:
SilverLining Location SetLongitude and SetLatitude accept degrees and not radians. Removed degrees to radians conversion.


gwaldron:
This fix is proper. But we also need to fix the problem that clouds are improperly lit when applying this fix. So I am holding off while we look into that. Thanks.

gwaldron:
Applied. Thanks.

PR #625, opened 2015-07-09
Changed injection point for heightmap fragment shader

leadcoder:
Changed injection point for heightmap fragment shader to avoid that heightmap get lit by the LOCATION_FRAGMENT_LIGHTING step.


gwaldron:
Not having used the heightmap, I have a question -- do the shaders need to inherit anything from the terrain engine at all? If not we could just call VirtualProgram::setInheritShaders(false), or even just use a simple osg::Program instead. Thoughts?

leadcoder:
Not sure If it’s possible but it use the raw elevation stored by the TileModelCompiler in osg::Drawable::ATTRIBUTE_6 
The purpose of the heightmap camera is to render the terrain from top down to get the terrain heightmap around current camera position so I suspect that the terrain rendering step should be done much like the normal terrain rendering to support lod blending etc.

gwaldron:
OK. That is just a vertex attribute binding; it will be available with or without the MP terrain shaders. What you would lose is something like LOD blending or vertical scaling -- anything that moves verts on the GPU. IIRC the heightmap's main function is to adjust transparency as the ocean approaches the shoreline so whether terrain shaders are critical I'm not sure.

PR #626, opened 2015-07-09
Option added to modulate image layer in the splat extention

leadcoder:
Environment variable option added to the splat extention to include imagery in the splat fragment shader. This way the we get some color variation to break up the tiling effects from the detail textures.

No imagery
splat02

Readymap imagery modulated
splat01


west036:
Could not work properly in TMS imagelayer of Mercator projection(EPSG:900913),Did I miss something?

gwaldron:
leadcoder, rather than accept this PR, we opted to re-enable the missing support for attentuation distance in the image layer. That way you can use an image layer as you normally would in combination with splatting. See the new "splat-with-imagery.earth" sample.

PR #628, opened 2015-07-14
KML fixes

AlexBobkov:
1) Style was passed to KML_Geometry by const reference. Then KML_Geometry modified the style copy. So AltitudySymbol and ModelSymbols were completely ignored during geometry creation step.

I've changed reference to non-const reference to take into account the style changes.

2) Default text symbol was not added to style for PlaceNode.
```
if ( !text && cx._options->defaultTextSymbol().valid() )
    text = cx._options->defaultTextSymbol().get();
```

3) I suggest to use UTF8 as default encoding for KML files.
```
text = style.getOrCreate<TextSymbol>();
text->encoding() = TextSymbol::ENCODING_UTF8;
```


gwaldron:
Thanks Alex

PR #644, opened 2015-08-27
Forced include headers for Qt5 moc on files without extension

podsvirov:
Note from moc documentation:
-f[file] - Force the generation of an #include statement in the output.
             This is the default for header files whose extension starts
             with H or h.


podsvirov:
Hi guys!
Thank you for your trust. For many months I have been using this fix for themselves to build with Qt5 for Windows.

A small digression.
I want to offer my project to look at the "DaD's House":

http://dad.podsvirov.pro

In it I collect a lot of interesting modules.
Maybe someone, it will also be interesting.
The 64-bit version of the testing installer contain osgEarth 2.7

PR #653, opened 2015-10-02
DateTimeRange class and KML fixes

AlexBobkov:
Solved issues in this thread: http://forum.osgearth.org/KML-Loader-td7586675.html


AlexBobkov:
I've added time animation capability according to this topic: http://forum.osgearth.org/Time-animated-annotations-td7588166.html

PR #660, opened 2015-11-02
Build fix against OSG-3.4.0 and OSG-3.5.0

robertosfield:
To get osgearth to build against changes to OSG-3.4.0 and OSG-3.5.0 I have had to introduce a couple of changes to use the new OSG API's.  Each new bit of API is guarded by the appropriate OSG version check.


scrawl:
Shouldn't the check say OSG_VERSION_GREATER_OR_EQUAL(3,5,0) rather than OSG_VERSION_GREATER_OR_EQUAL(2,5,0) ?

robertosfield:
Oooopss.... You are right it should have been 3,4,0 (for silverlining code)
and 3,5,0 for ElevationQuery.  I have now checked this fix in and
simplified the check so it does check for the existing and the
OSG_VERSION_GREATER_OR_EQUAL as the rest of osgEarth doesn't bother with
this check.

jasonbeverage:
Thanks Robert!

PR #662, opened 2015-11-05
Stroke Texturing

rmk177:
I add some stroke texturing functionality. Image of road is from http://seamless-pixels.blogspot.ru and I add new boston-gpu_tex.earth file to add road texuring to osgearth. 
simply add stroke-texture: ../data/road3e.jpg; to have roads textured.


gwaldron:
Thanks. The original intention was to tie it into the resource library "skin" framework in the same way as the extrude filter. So we need to think about this a bit more to make sure things are consistent.

jasonbeverage:
Thanks for this, but as Glenn said I think the way we would implement this is with the skin framework.

PR #663, opened 2015-11-06
Link X11, fixes #657

### LINKED ISSUE #657: undefined reference to symbol 'XInitThreads'
When trying to build osgEarth 2.6 or 2.7 on Ubuntu 14.04 I'm running into:

```
[ 97%] Linking CXX executable ../../../bin/osgearth_qtd
/usr/bin/ld: CMakeFiles/application_osgearth_qt.dir/osgearth_qt.cpp.o: undefined reference to symbol 'XInitThreads'
//usr/lib/x86_64-linux-gnu/libX11.so.6: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
```

I've found this fix: http://permalink.gmane.org/gmane.comp.gis.grass.pkg.devel/30390

Can this be merged upstream?

jhasse:
This might break building on Windows. Should I use `if(UNIX)` for that?

### DIFF
--- a/src/applications/osgearth_demo/CMakeLists.txt
+++ b/src/applications/osgearth_demo/CMakeLists.txt
@@ -16,6 +16,7 @@ ELSE()
         ${QT_QTCORE_LIBRARY}
         ${QT_QTGUI_LIBRARY}
         ${QT_QTOPENGL_LIBRARY}
+        X11
     )
 ENDIF()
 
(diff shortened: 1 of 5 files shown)


gwaldron:
I'm not convinced this is necessary; we have many users that build night on Linux without issue. Perhaps the problem lies elsewhere in your CMake configuration?

PR #665, opened 2015-12-01
Fix gcc -Wreorder warnings

arnaudgelas:
This patch fixes several hundred warnings when using gcc


jasonbeverage:
Thanks for this, but we took too long to merge it and it's too many changes to pull in now that there are conflicts.

PR #674, opened 2015-12-09
Refactoring conditional directives.

wmarquesr:
A suggestion to compile entire statements and expressions, as suggested by code style guidelines from the Linux Kernel and practitioners.

* [Linux kernel coding style](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/CodingStyle#n892)
* [Living in the #ifdef Hell](https://www.cqse.eu/en/blog/living-in-the-ifdef-hell/)

It might improve code understanding, maintainability and error-proneness.


gwaldron:
"duktape.c" is a source file from an external project; we include it in the distribution for convenience but we did not author it. (http://duktape.org/) So it would be more appropriate to submit changes to that project directly.

On a side note: I could not even spot your ifdef changes because of the myriad formatting changes in the diff. If you're going to change the formatting, please do so in a separate commit. Thanks.

PR #675, opened 2015-12-15
Fix feature annotation messy code on windows platform

raindust:


gwaldron:
Won't this cause problems when an attribute value uses a different code page?

remoe:
I think this is not needed. We work with windows platform and don't have issues with it.

PR #708, opened 2016-04-11
Rotation on LabelNode and PlaceNode

JD31:
Two new attributes on TextSymbol:
- text-rotation: absolute on-screen rotation
- text-geographic-course: compute the on-screen rotation for following
  a fixed course on ground (use for naming rivers or roads)

Can be tested with annotation.earth (see the tesselated line)

### DIFF
--- a/src/osgEarth/Decluttering
+++ b/src/osgEarth/Decluttering
@@ -55,6 +55,7 @@ namespace osgEarth
     public:
         float      _priority;
         osg::Vec2s _pixelOffset;
+        float      _localRotationRad;
     public:
         virtual ~LayoutData() { }
     };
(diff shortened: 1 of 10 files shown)


JD31:
Thank you for merging. I hope it will be usefull for others.
But on last commit you've modified updateLayoutData method (labelnode and placenode) and the call to setUserData is now out of the "if (!_dataLayout.valid())" block which i think is very not optimal.
I can see that the indentation was not correct and maybe that is reason why the source has been adapted...
Can you move back "setUserData" into "if (!_dataLayout.valid())" block please?

gwaldron:
Even if the dataLayout is already valid, we still need to call setUserData() because the list of Drawables might have changed. For example, LabelNode::setStyle() removes all the Drawables and recreated them, and then calls updateLayoutData().

PR #710, opened 2016-04-14
Update OsgEarthMacroUtils.cmake

haifeng60902:
add  lib perfix  such as  earth29-earth.dll , put  earth plugin to local osgPlugin-3.4.0 directory.


gwaldron:
Please explain.

haifeng60902:
I do this is to avoid in Windows platform for the use of multiple versions of DLL Hell. Plug-in is placed in a directory in osg osgPlugin - 3 X, I think it is more advantageous to the coexistence of multiple versions. Because, of course, there is nothing in the main file is similar to osg version number, so I adopted osgearth version number, if you think it reasonable behind we can separate the DLL version number specified.

gwaldron:
Plugin DLLs already go into an osgPlugins-3.x.x folder if you set OSG_PLUGINS in the CMAKE configuration. Better yet, you can set OSG_VERSION_EXE and it will figure out the folder name for you.

PR #718, opened 2016-04-26
Allow bounding box drawing around label

JD31:
New symbol added BboxSymbol.
Example:
					text-bbox-fill: #00FF0033;
					text-bbox-margin: 3;
					text-bbox-border: #FFFFFFFF;
					text-bbox-border-width: 1;
An example has been added in annotation.earth file.


gwaldron:
The bbox symbol might better belong in the TextSymbol, but we can make that change later if it makes sense.

JD31:
OK. You are right.
I have also a small issue with the implementation I did:
As soon as the centroid of the bounding box geometry moves out of the screen it disappears. That's not the case with the Text drawable and thus I can have the text without the decoration when close the screen border. Do you know where i can search? (bounding box seems OK and I don't really understand what is so different between both drawables...)

gwaldron:
I can reproduce the problem. Nothing immediately comes to mind, but we will look into it today.

jasonbeverage:
I just pushed a fix for this in 8bb648fa62cd801ebd0cc7eecdea4736d7f06b32.  The text drawable for a LabelNode has culling explicitly disabled with setCullingActive(false).  I've disabled culling of the bounding box drawable in the same way and it appears to work correctly now.  The reason that culling needs to be disabled is b/c the drawables are eventually rendered using the screen space layout, so their world space bounding boxes aren't necessarily what is finally rendered.

JD31:
Thank you for analysis.
However I don't see the fix in the given commit, nor in the master...

jasonbeverage:
Sorry about that, I committed, typed git push and left for a trip.  Didn't
see that I needed to pull before my push :)  It should be in there now.

PR #719, opened 2016-04-28
Fix unsetting options of CacheBin write in URI class

raindust:
@gwaldron added options to most caching functions a month ago, whoever the options did not pass to CacheBin write method.


gwaldron:
Thanks.

PR #724, opened 2016-05-18
Removed setCullingActive(false) to fix compile error

mahisorn:
I am very new to osgEarth and found that setCullingActive(false) will prevent the master branch to compile.


mahisorn:
I think I should put this in Issues instead of pull request.

JD31:
The min Osg version required is 3.4.0.
I am posting a merge request with the necessary pre-processing.

mahisorn:
Thanks for the information.

PR #725, opened 2016-05-18
Allow setCullingActive call only with OSG >= 3.4.0

JD31:
Ensure that the OSG version is compatible with setCullingActive() method like it is done for Texts


gwaldron:
Merging this will break the functionality -- and the next release will require OSG 3.4+ anyway. So perhaps it's best to forgo this?

JD31:
I refer to:
https://github.com/gwaldron/osgearth/pull/724

OsgEarth will not compile without this patch (but it will not break the functionnality if OSG is >= 3.4).
Morover it is homogenous with createTextDrawable from AnnotationUtils.

It's up to you now!

PR #733, opened 2016-06-01
Asan

arnaudgelas:


arnaudgelas:
@gwaldron Since the compilation error has already been fixed, so would you mind cherry-picking the second one? or should I create another pull request only with the asan fix?
Thank you very much

jasonbeverage:
Can you explain what error adding the assignment operator actually fixes?  We don't have explicit assignment operators for the other Options classes and haven't ran into any issues that I'm aware of.  If it truly points to a real issue then it would make sense that the other Options classes need explicit assignment operators too.

arnaudgelas:
Trajce (@trajcenikolovnick) fixed a similar bug in #697  in CachePolicy. Adding this implementation of operator= avoid crashes due implicit copy conversion...

This commit is the minimal fix to make things work when using osgearth in our application. Indeed adding copy constructor with similar implementation may fix similar bugs. I'll update the pull request with the copy constructor.

arnaudgelas:
If needed, I can squash these 3 commits into a single one, cherry-pick and create another pull request. Let me know, what is best for you, please!

arnaudgelas:
see #734

PR #735, opened 2016-06-02
EarthManipulator terrain avoidance minimum distance setting added, no…

emminizer:
… longer using minimum focal distance for terrain avoidance.

### DIFF
--- a/src/osgEarthUtil/EarthManipulator
+++ b/src/osgEarthUtil/EarthManipulator
@@ -37,11 +37,11 @@ namespace osgEarth { namespace Util
 {
     using namespace osgEarth;
 
-    /** 
+    /**
      * A programmable manipulator suitable for use with geospatial terrains.
      *
      * You can use the "Settings" class to define custom input device bindings
-     * and navigation parameters. Create one or more of these and call 
+     * and navigation parameters. Create one or more of these and call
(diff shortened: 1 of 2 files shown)


jasonbeverage:
Thanks @emminizer! This looks fine to merge but what problem does this solve exactly.  Were you having issues?

emminizer:
This resolves a specific scenario for us where we were receiving reports in our osgEarth-based app that terrain avoidance was not functioning.  It was not functioning because we use a negative "minimal focus distance" to allow for zooming through models.  We do this to support a feature called Cockpit mode and it works great.  Imagine being tethered to an aircraft model, following its yaw, pitch, and roll, zoomed forward so you're sitting on the nose.

But by setting a negative focus distance, it messes up the terrain avoidance minimum distance, which uses the same variable.  We've seen lots of funky behavior with terrain avoidance doing something "close to right" but not quite right.  I finally tracked it down to the reuse of this variable.  The only documentation that I could find that this variable was used was from Glenn on our internal help desk, who mentioned this on the same issue that this change resolves (SIM-4453).

In short, this lets us independently vary the minimum "zoom" value and keep it distinct from the "minimum distance from terrain" variable.  Though they're both minimums, they represent two different values both conceptually, and (in this patch) in the code as well.

PR #741, opened 2016-06-09
MapNode and EarthManipulator no longer crash when terrain driver (e.g…

emminizer:
…. engine_mp) is not found.

Was getting a crash when the MP driver was explicitly used in configurations when the engine_mp plug-in was not found (typically due to end user misconfiguration).  This resolves the hard crash by adding NULL checks.

### DIFF
--- a/src/osgEarth/MapNode.cpp
+++ b/src/osgEarth/MapNode.cpp
@@ -48,7 +48,7 @@ using namespace osgEarth;
 
 //---------------------------------------------------------------------------
 
-namespace 
+namespace
 {
     // adapter that lets MapNode listen to Map events
     struct MapNodeMapCallbackProxy : public MapCallback
@@ -101,13 +101,13 @@ public:
       }
 
       virtual void apply(osg::PagedLOD& node)
-      {        
-          //The PagedLOD node will contain two filenames, the first is empty and is the actual geometry of the          
+      {
+          //The PagedLOD node will contain two filenames, the first is empty and is the actual geometry of the
           //tile and the second is the filename of the next tile.
           if (node.getNumFileNames() > 1)
           {
               //Get the child filename
-              const std::string &filename = node.getFileName(1);              
+              const std::string &filename = node.getFileName(1);
               if (osgEarth::Registry::instance()->isBlacklisted(filename))
               {
                   //If the tile is blacklisted, we set the actual geometry, child 0, to always display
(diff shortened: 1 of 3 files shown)


src/osgEarth/MapNode.cpp
```diff
@@ -388,14 +388,22 @@ MapNode::init()
...
+                _map->getSRS()->getEllipsoid()->getRadiusEquator(),
+                _map->getSRS()->getEllipsoid()->getRadiusPolar());
+            stateset->addUniform( new osg::Uniform("oe_ellipsoidFrameInverse", osg::Vec3f(ellipFrameInverse)) );
+
+            osg::Vec3d ellipFrame = osg::componentDivide(osg::Vec3d(1.0,1.0,1.0), ellipFrameInverse);
+            stateset->addUniform( new osg::Uniform("oe_ellipsoidFrame", osg::Vec3f(ellipFrame)) );
+        }
+        else
+        {
+            stateset->addUniform( new osg::Uniform("oe_ellipsoidFrameInverse", osg::Vec3f()) );
```
emminizer: I am unsure if this is the best default value for oe_ellipsoidFrameInverse (and oe_ellipsoidFrame).  Would osg::Vec3f(1.f, 1.f, 1.f) be better?

jasonbeverage:
Is this to prevent a crash when the osgdb_osgearth_engine_mp.dll is not accessible at all, like it's not in your path?  What would osgEarth do if it can't find the terrain engine at all?

emminizer:
> Is this to prevent a crash when the osgdb_osgearth_engine_mp.dll is not accessible at all, like it's not in your path?

Right, osgdb_osgearth_engine_mp.dll is not accessible at all.  This can happen if the end user developing an app messed up on their paths, or during configuration, or during a bad deployment.  We develop an SDK that uses osgEarth as a baseline and we get a surprising number of issues from end user developers that boil down to bad paths and not being able to load OSG plug-ins because of this.

> What would osgEarth do if it can't find the terrain engine at all?

The only thing I would expect (perhaps "hope for") is that it not crash.  Beyond that, I'd expect that it couldn't do anything useful.  Most of the code already looks quite well protected against this, there were only a small handful of places I was able to spot in a "normal" run where pointer validity wasn't checked.

In our SDK we try to detect this condition and print debugging info to the end user (e.g. osgDB::Registry::getLibraryFilePathList() search path and the MP DLL name).  When the user has a misconfigured path, what we're seeing is an error from osgEarth about MP, then a crash (due to unprotected NULL accesses).  This PR simply prevents the NULL crashes in the cases I spotted.

PR #768, opened 2016-08-11
Render bin is now a setting for TritonOptions.  Removed setRenderingH…

emminizer:
…int() code that caused a problem with displaying the surface.


emminizer:
Prior to this PR change, I was getting a fully transparent ocean surface that still wrote to the depth buffer (and obscured scene entities), due to the move of setRenderHint() from the Drawable to the Node.  Updated by following the pattern of simple_ocean of passing in a render bin and defaulting it to the same bin that simple_ocean uses.

PR #772, opened 2016-08-16
SilverLining sky now correctly reacts to SkyNode's minimum ambient an…

emminizer:
…d date changes.


emminizer:
Note that there are still outstanding issues where _contexts might need to be used instead of _children.  It wasn't obvious to me how to correctly address this, so I did not attempt to do so.

This patch instead just addresses the Sky node observer method, ensuring that it iterates through the list of contexts properly.  So it doesn't fix the whole problem, but it makes it better than it was.

emminizer:
I updated the diff so that the issue

PR #773, opened 2016-08-19
Triton: Now cleaning up terrain callback properly, preventing crash w…

emminizer:
…hen changing ocean models.


emminizer:
In our application, we let the user change the ocean model.  Swapping out from Triton leaves stale pointer in not-removed terrain callback.  This both removes the callback and protects the TritonDrawable in an observer_ptr.  This fixes stale memory access and prevents crashes.

Also added some safety to the _contextDirty buffered_object to prevent out-of-bounds access.

PR #775, opened 2016-08-23
SilverLining: Atmospheric Turbidity, Light Pollution added.  Integrat…

emminizer:
…ed callback for cloud millisecond timing.


emminizer:
The millisecond time is required in order to stop cloud motion on demand, or to speed up or slow down cloud animation.  It was difficult to deal with memory management by putting this into the SilverLiningAPIWrapper code, so instead I added the callback to the list of Callbacks.  The default behavior remains unchanged.

Added turbidity and light pollution accessors since I was already in similar code.

PR #780, opened 2016-08-25
SilverLining: Initialize Minimum Ambient, if set prior to context nod…

emminizer:
…e creation.


emminizer:
The context node is created on the next UPDATE traversal after sky is attached.  Calls to SkyNode::setMinimumAmbient() therefore get ignored until the first UPDATE traversal after attaching.

PR #796, opened 2016-09-09
CMakeModule: Modify find gdal order

KangLin:


jasonbeverage:
Thanks for the PR, can you explain why this is needed?  The FindGDAL.cmake hasn't changed in 5 years and as far as I know we've had no issues related to it.

KangLin:
I use vs2015 in msys2 bash environment, but it finds gdal.lib. Then, it can not find gdal.lib dependent libraries (for example: odbc, etc.). However, gdal_i.lib containing these dependent information. So the gdal_i to the front, so FindGDAL.cmake to find it.

PR #797, opened 2016-09-09
GeoPositionNode can now be configured to ignore setMapNode.  Some ins…

emminizer:
…tances require having a NULL map.


emminizer:
This fixes a problem we are experiencing in our software where we create LocalGeometryNode instances with an intentionally NULL map that are attached to nodes in our scene.  When we reload the map, the MapNodeObserver comes in and, trying to be helpful, assigns the map node.  This breaks the display.  The use case is an extruded relative line overlay.

We can possibly address this in our own code base however, so if this is not a solution you want to see, feel free to close the PR and we can look for an application level solution.

Issue #800, opened 2016-09-14
Including files with xml tag <xi:include> only includes the first tag


rhabacker:
Loading an earth file including other files having multiple top level tags like <image> or <model> only displays the first tag. See the example:

test.earth
-------------------------------------------------------------------
```
<map name="test" type="projected" version="2">
      <xi:include href="part.xml"/>
</map>
```

-------------------------------------------------------------------
part.xml
-------------------------------------------------------------------
```
<image name="readymap_imagery" driver="tms">
      <url>http://readymap.org/readymap/tiles/1.0.0/7/</url>
</image>
<elevation name="readymap_elevation" driver="tms">
       <url>http://readymap.org/readymap/tiles/1.0.0/116/</url>
</elevation>
```
-------------------------------------------------------------------

Running 

OSGEARTH_NOTIFY_LEVEL=DEBUG osgearth_viewer test.earth 

shows that only the entry named "readymap_imagery" is included.

jasonbeverage:
That's a known limitation of the xml include spec, and the reason we wrote the [template driver](https://github.com/gwaldron/osgearth/tree/master/src/osgEarthDrivers/template)

Check out the [readymap_template.earth](https://github.com/gwaldron/osgearth/blob/master/tests/readymap_template.earth) example instead.  

You have to append .template to the earth file when you load it to invoke the psuedoloader like this:

osgearth_viewer readymap_template.earth.template

rhabacker:
Thanks for this hint. The irritating thing is that it works for viewpoints https://github.com/gwaldron/osgearth/blob/master/tests/viewpoints.xml, but not for other. If not available I suggest to add a <group> tag or similar to be able to group several tags together, which could then be placed into a seperate xml file. 
The additional benefit is, that this tag could have an "enabled" attribute to switch on/off the whole group similar to the "enabled" attribute of the <image> or <model> tag.

jasonbeverage:
It works for viewpoints b/c that file is wrapped in a higher level
<viewpoints> block, it's not just a list of <viewpoint> elements.

rhabacker:
Yes and therefore I suggested the addition of a common block like `<group>` for example 

```
<group>
  <image>
  </image>
</group>
```
or 
```
<group>
  <model>
  </model>
</group>
```

where `<group>` is optional in the main file and mandatory in the sub xml file.

gwaldron:
I commented on this in the PR, but I will repeat here for convenience.

[...]

PR #801, opened 2016-09-15
Remove limitation of only supporting one earth file element in included xml files.

rhabacker:
Included xml files via xml <xi:include> tag currently supports only
one earth file element because xml files require a top level xml
tag. This limitation is fixed by adding support for an xml <group>
element as top level xml tag.

see bug #800


gwaldron:
Thanks for the PR. At this time we don't think this the right solution. Generally speaking (in GIS tools e.g.) a Group Layer connotes the ability to operate on the layer as a Group in all capacities (toggling visibility, changing render order, and so on), while your approach is only a deserialization semantic. It also doesn't persist in the Map model, so the group will not be present if you re-serialize the same MapNode. 

We have contemplated the idea of Group layers quite a few times over the years but have not implemented them because the concept does carry these expectations.

PR #811, opened 2016-10-05
Sky::getMinimumAmbient: No longer returning reference to temporary.

emminizer:
I created an error in a previous pull request.  This fixes that error.  Sorry.


gwaldron:
Already fixed. Thanks

PR #812, opened 2016-10-05
Capabilities: getGLSLVersionInt now rounds to avoid precision bug

emminizer:


emminizer:
Discovered this on a machine that has GLSL 3.3.  getGLSLVersionInt() was returning 329, but the string was returning 3.3

PR #823, opened 2016-11-07
Implement Cache::clear()

arnaudgelas:
To this end, PerObjectRefMap::clear needed to be implemented too (simple
call to clear method of std::map<>)


gwaldron:
Your implementation of Cache::clear is incorrect. Clearing the collection of CacheBin objects does not actually delete any records in those underlying cache bins.

PR #826, opened 2016-11-11
Change the _models order in the MapNode

njuliujl:
The models should be visited before the OverlayDecorator, during a
culling traversal, so that the cullOverlayGroup method in DrapingTechnique
or ClampingTechnique may work correctly.

### DIFF
--- a/src/osgEarth/MapNode.cpp
+++ b/src/osgEarth/MapNode.cpp
@@ -276,6 +276,15 @@ MapNode::init()
     // TODO: not sure why we call this here
     _map->setGlobalOptions( local_options.get() );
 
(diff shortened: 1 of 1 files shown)


gwaldron:
What problem are you trying to solve here? Are you seeing a bug?

njuliujl:
When I was tracking the program, I found that the OverlayDecorator is visited before the models that uses draping technique. 
I think the drapable models should be visited first and be collected in the DrapingCullSet, so that the DrapableNodes  are traversed later in the DrapingCamera which is done when the  cullOverlayGroup method of DrapingTechnique is invoked.

PR #831, opened 2016-11-17
Rex engine crash fix; check for NULL.

emminizer:
_terrain can be NULL in addTileLayer in some cases.


gwaldron:
Dan, thanks. I made my own fix to this and pushed it.

PR #835, opened 2016-12-01
ReaderWriterWMS: Locale independent bounds serialization

podsvirov:
BBOX of eight numbers? It's true for locale with desimal point is the `,`.

Use the `sprintf` function is not safe here. The result depends on the locale.

The result is an improper URL request which can not be recognized cartographic server.

This request is a patch for a quick solution to the problem. Probably it is necessary to improve the implementation, but I have no time to do it. I just want to draw attention to the problem.

### DIFF
--- a/src/osgEarthDrivers/wms/ReaderWriterWMS.cpp
+++ b/src/osgEarthDrivers/wms/ReaderWriterWMS.cpp
@@ -480,8 +480,15 @@ public:
         double minx, miny, maxx, maxy;
         key.getExtent().getBounds( minx, miny, maxx, maxy);
         
+        // Convert numbers only in "C" locale (decimal point is dot)
+        // Store current numeric locale
+        char *cur_lc_numeric = setlocale(LC_NUMERIC, "C");
+
         char buf[2048];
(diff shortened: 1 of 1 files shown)


gwaldron:
I agree that this is not a thread-safe solution; will leave it open until we come up with something.

podsvirov:
@gwaldron well. Let's wait for the results.

podsvirov:
Hello @gwaldron, please point to solution.

boitumeloruf:
Why was this merge request closed without being merged? This issue of an improper URL request is still occurring. Please fix.

gwaldron:
Because in 3 years there has been no attempt to update the solution. Feel free to do so.

boitumeloruf:
I have integrated the changes of podsvirov into the current head and created a new merge request: https://github.com/gwaldron/osgearth/pull/1415

I have tested the changes on a machine with a german locale. The changes fixed the problem.

tompollok:
This PR works for me aswell @gwaldron whats blocking the pr provided by @boitumeloruf

PR #837, opened 2016-12-05
VAO support

tomhog:
Hey Glenn

Here are the changes merged into master. It's actually just adding a vertex buffer object to for the texture coords and switching VAOs on.

Basically you can bind and unbind vbos while a vao is bound your just kind of losing the benefit of vaos. The issues where being caused by the fact the texture coords where sharing their vbos, altering that in vao was the issue.

Previously I did have to override createVertexArrayState in MPGeometry as their weren't enough array dispatchers allocated for the other texture cord channels, but that crash seems to have disappeared.

Anyhow it runs on my mac with a GL3 context.

### DIFF
--- a/src/applications/osgearth_viewer/osgearth_viewer.cpp
+++ b/src/applications/osgearth_viewer/osgearth_viewer.cpp
@@ -52,6 +52,8 @@ main(int argc, char** argv)
 
     float vfov = -1.0f;
     arguments.read("--vfov", vfov);
+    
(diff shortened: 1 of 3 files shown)


gwaldron:
Tom, the PR failed the CI build (see above); please review, thanks

tomhog:
Ok should have fixed that, i had added the setVertexBufferHint call into the viewer example and forgot to protect it.

I think it's rebuilding now. That auto build thing is very cool.

PR #848, opened 2017-01-13
[DON'T MERGE] support geos 3.6.1

filnet:
For future reference.

Compiles but has not been tested yet...

Apparenlty macos brew now offers geos 3.6.1

### DIFF
--- a/src/osgEarthSymbology/GEOS.cpp
+++ b/src/osgEarthSymbology/GEOS.cpp
@@ -212,7 +212,7 @@ GEOSContext::GEOSContext()
     geos::geom::PrecisionModel* pm = new geos::geom::PrecisionModel(geom::PrecisionModel::FLOATING);
 
     // Factory will clone the PM
-    _factory = new geos::geom::GeometryFactory( pm );
+    _factory = geos::geom::GeometryFactory::create( pm );
 
     // Delete the template.
     delete pm;
@@ -220,7 +220,6 @@ GEOSContext::GEOSContext()
 
 GEOSContext::~GEOSContext()
 {
-    delete _factory;
 }
 
 geom::Geometry*
@@ -229,12 +228,7 @@ GEOSContext::importGeometry(const Symbology::Geometry* input)
     geom::Geometry* output = 0L;
     if ( input && input->isValid() )
     {
-        output = import( input, _factory );
-
-        // if output is ok, it will have a pointer to f. this is probably a leak.
-        // TODO: Check whether this is a leak!! -gw
-        //if ( !output )
-        //    delete f;
+        output = import( input, _factory.get() );
     }
     return output;
 }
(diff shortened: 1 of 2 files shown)


filnet:
Note that the PR is against branch 2.7

willysr:
Tested with osgearth 2.8 and it compiles.
+1 of merging it.

filnet:
msys2 now also offers geso 3.6.1

outruner:
Hello after applying this patch on 2.8 I could successfuly compile with  geos 3.6.1.
However  I have a crash on osgearth_viewer with following earth : feature_country_boundaries.earth.
it seems that this GeometryFactory is null for some reason. Can anyone reproduce this issue?

outruner:
Hello after applying a little modification to the patch I could succcessfuly run the sample. Please read my comment on the patch for more informations (unique_ptr should be passed by reference and not by value). regards.

filnet:
I pushed a fix for the crash.

src/osgEarthSymbology/GEOS.cpp
```diff
@@ -66,7 +66,7 @@ namespace
     }
 
     geom::Geometry*
-    import( const Symbology::Geometry* input, const geom::GeometryFactory* f )
+    import( const Symbology::Geometry* input, const geom::GeometryFactory::unique_ptr f )
```
filnet: i reverted to passing a pointer to the factory instead of a unique_ptr

outruner:
thanks !

filnet:
Tried the fix with master.

osgearth_viewer.exe src/osgearth-osgearth-2.8/tests/feature_country_boundaries.earth works with the proposed fix, but there is a segfault when exiting (on msys2 at least).

I also see a bunch of message like this :

error: linking with uncompiled shadererror: linking with uncompiled shadererror: linking with uncompiled shadererror: linking with uncompiled shader
VERTEX glCompileShader "oe_mp_vertModel" FAILED
VERTEX Shader "oe_mp_vertModel" infolog:
0:1(10): error: the compatibility profile is not supported
0:1(10): error: GLSL 3.30 is not supported. Supported versions are: 1.10, 1.20, 1.30, 1.00 ES, and 3.00 ES

outruner:
in release mode there is no crash (fork of 2.8), compiler msvc2015. However in debug mode I also have a crash on exit (mutex.lock issue) , stacktrace attached. I did not see any shader or opengl error in both modes.
crash_osg

filnet:
The GLSL problem is probably related to msys2.

I get the segfault with a release build but i am on msys2 not mscv.
I can't say if the mutex related segfault is related to geos (?).

filnet:
osgearth_viewer with simple_model.earth also exhibits the segfault on exit.

So, I guess, it is safe to say that the segfault is not due to geos upgrade.

manisandro:
Patch based on which preserves GEOS < 3.6.0 compatibility: https://github.com/manisandro/osgearth/commit/9cb5029b86a7a51ce367ffe6a190f6a26f7f418d

filnet:
@manisandro best would be to turn your patch into a PR on master

manisandro:
Done, I suppose this one should be closed.

filnet:
superseded by https://github.com/gwaldron/osgearth/pull/899

PR #850, opened 2017-01-18
Add USE_QT4 to allow building with Qt4 when Qt5 also present

cjmayo:
I'm building osgEarth to support the Globe plugin for QGIS. QGIS stable is still recommending Qt4 at this time but I also have Qt5 installed. Currently osgEarth will automatically build for Qt5 for me.

This is my quick solution, adding a `USE_QT4` option. Submitting here in case it is useful for others.

Without USE_QT4 it will still build for Qt5 if installed. It won't automatically fallback from trying Qt5 to try Qt4 - arguably that's an improvement.
Happy to change the logic, create a OSGEARTH_USE_QT4? etc. Possibly CMake should fail if OSGEARTH_USE_QT is set and no Qt is found.


manisandro:
Globe maintainer here. I'd like to go further with this and actually allow side-by-side installation of the qt4 and qt5 variants (to allow developing both qgis-2.x and qgis-3.x without library chaos). I've got a patch which, in addition to allowing to choose the Qt version via `USE_QT5={true/false}`, changes the names of the osgEarthQt library and the qt applications to allow installing both variants without conflicts. The patch is here: https://github.com/manisandro/osgearth/commit/c90d28be49ecc3fff2133320be8d3f5f38685c29
@cjmayo  Shall I open a new PR or do you want to integrate these changes into this PR?

A similar patch I've proposed to osgQt has been merged: https://github.com/openscenegraph/osgQt/pull/3

gwaldron:
You can integrate the changes, but please take care with the versioning. osgEarth must be able to build against OSG 3.4 up through the master branch.

cjmayo:
@manisandro I've tested (osgEarth 2.8.0 with OpenSceneGraph 3.4.0) manisandro@80901d4 for my case of both Qt4 and Qt5 installed and only building for Qt4 and as it does that and more happy to close this one.

manisandro:
Just for info, I'm waiting for https://github.com/openscenegraph/osgQt/pull/4 and then I'll prepare the PR for osgearth.

PR #852, opened 2017-01-20
Re added createVertexArrayStates fix

tomhog:
Hi Glenn

Remo ran into an issue I was previously having with VAO on OSX but I thought it had been fixed within osg. Turns out it hadn't

So the change to MPGeometry is required because on line 239 of MPGeometry.cpp a texture coord unit is used that wasn't part of the geometry texture arrays. This means no vertex array dispatcher was created for it so it crashes.

I feel like it should nearly always crash, but it seems fine in some cases and like I said it stopped crashing on me, but a .earth file with no layers causes it to happen still.

The fix is pretty ugly I think, but it works and I've made it only apply when VOAs are in use as no one else seems to have the issue.

I also fixed a small compiler error in HTTPClient.cpp

My topic-vao-osx branch was merged with latest master before I did these changes

### DIFF
--- a/src/osgEarth/HTTPClient.cpp
+++ b/src/osgEarth/HTTPClient.cpp
@@ -1311,7 +1311,7 @@ HTTPClient::doGet(const HTTPRequest&    request,
 
     METRIC_END("HTTPClient::doGet", 2,
                "response_code", toString<int>(response.getCode()).c_str(),
-               "canceled", toString<bool>(response.isCancelled()));
+               "canceled", toString<bool>(response.isCancelled()).c_str());
 
     return response;
 }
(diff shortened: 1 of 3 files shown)


gwaldron:
Thanks Tom; you need to ifdef this so it will still compile under OSG 3.4.0

tomhog:
Sorry about that, forgot you had the funky auto test feature setup. I've made the changes. Looks like it's building.

Does that test happen automatically when I do a pull request or is it something you have to trigger?

gwaldron:
It's automatic. TravisCI monitors the GitHub repo for commits, PRs and fires off a new check when it sees one.

tomhog:
Thanks Glenn

Issue #860, opened 2017-02-01
GL_QUAD's are deprecated on GL3.x


remoe:
Some code use GL_QUAD's. For example: controls.cpp. One can port such code:

```c++
_geom->addPrimitiveSet( new osg::DrawArrays( GL_QUADS, 0, 4 ) );
```

to

```c++
GLushort indices[] = {0,1,2,0,2,3};
_geom->addPrimitiveSet( new osg::DrawElementsUShort( GL_TRIANGLES, 6, indices ) );
```

After this it, the controls also run on MacOSX GL3.

PR #876, opened 2017-02-17
Fix: #860 GL_QUAD's are deprecated on GL3.x  

remoe:


gwaldron:
Remo, why expressly enable the VAOs on instead of using the osg::DisplaySettings flag?

remoe:
Glenn, i've changed it.

PR #878, opened 2017-02-27
2.8 - GDAL static linking support and Visual Studio build fixes

Pianotorious:
Support linking GDAL statically (this requires that we initialized the GDAL and OGR drivers for each shared library that uses GDAL).
Added /FORCE:MULTIPLE for MSVC to fix several projects that weren't building on VS2013 because of multiple definition errors given for ostream.  (This is a fairly well-known workaround for link errors in osgDB with visual studio).


Pianotorious:
I realize there's a bit of copy-paste in how I set up the GDAL initializers -- there are multiple InitGDAL.cpp files with basically the same code in them.  I did this because IIRC CMake is not keen on you adding source files from parent directories, meaning I didn't really have a way of sharing the same .cpp file between multiple projects.  Let me know if you know a better way of doing it...

gwaldron:
OBE.

PR #879, opened 2017-02-28
Added conditional support for exp2 to fix MSVC 2010 and 2012 builds.

emminizer:
Added an ifdef around exp2 usage, which is not available on all compilers.  An alternate solution is to replace exp2 with pow(2.0, dL), but there are possibly performance or precision issues with that solution.


emminizer:
This is fixed in a different equally good or better way in master.  Should I discard / close this out?

gwaldron:
Fixed independently. Thanks.

PR #882, opened 2017-03-02
HorizonCullCallback: Optimize proxy node case to not fall back to own…

emminizer:
…er node position if proxy node exists and is not visible.


emminizer:
In a particular scenario with 10,000 nodes hitting the HCC all with proxy nodes, this improves the cull performance by 10-20% by relying on the is-visible for the configured proxy.

PR #890, opened 2017-03-22
Allow side-by-side installation of osgEarth compiled against Qt4 and Qt5

manisandro:
- Add the `DESIRED_QT_VERSION` CMake option to control whether osgEarthQt is compiled against Qt4 or Qt5 (following the convention of osgQt which also uses `DESIRED_QT_VERSION`)
- Allow conflict-free side-by-side installation of osgEarthQt compiled against Qt4 and Qt5 by renaming specifying Qt5 as appropriate in the libraries and examples compiled against Qt5, following the convention used by upstream Qt (i.e. libQt5Core):
  - osgEarthQt**5**
  - osgearth_package_qt**5**
  - osgearth_qt**5**_simple
  - osgearth_qt**5**_windows
- Preserve `<osgEarthQt/...>` include paths. Since headers are identical between osgEarthQt and osgEarthQt5, there is no point distinguishing them and breaking applications which `#include <osgEarthQt/...>`.
- Look for `osgQt5` for Qt5 build if OpenSceneGraph version is newer or equal 3.5.6.

See also https://github.com/openscenegraph/osgQt/pull/3, https://github.com/openscenegraph/osgQt/pull/4.


manisandro:
Ping?

jasonbeverage:
Thanks for this.  Like Robert we don't really have the QT expertise either, so thanks for taking over the maintenance of osgQT and providing patches.

manisandro:
Welcome, thanks for merging.

PR #894, opened 2017-03-28
Controls: LabelText compile fix for latest OSG 3.5.6

remoe:


remoe:
The latest OSG master doesn't compile without this fix.

PR #900, opened 2017-04-04
2.8

gedalia:
added nvtt based mipmapping in background thread.
names for nsight and debugging


gedalia:
It looks like this might be spewy/slow if nvtt isn't present, might want to cache the pointer the first time it's used.

PR #901, opened 2017-04-05
msys2: fix link error

filnet:
since upgrading to gcc 6.3.0 some plugins fail to link
because of undefined reference to `osgEarth::ProfileOptions::mergeConfig(osgEarth::Config const&)'

### DIFF
--- a/src/osgEarth/Profile
+++ b/src/osgEarth/Profile
@@ -75,10 +75,7 @@ namespace osgEarth
         Config getConfig() const;
 
     protected:
-        void mergeConfig( const Config& conf ) {
-            ConfigOptions::mergeConfig( conf );
-            fromConfig( conf );
-        }
+        virtual void mergeConfig( const Config& conf );
 
     private:
         void fromConfig( const Config& conf );
--- a/src/osgEarth/Profile.cpp
+++ b/src/osgEarth/Profile.cpp
@@ -56,6 +56,12 @@ _numTilesHighAtLod0( 1 )
     _namedProfile = namedProfile; // don't set above
 }
 
+ void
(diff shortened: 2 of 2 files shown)


filnet:
Full error sample:

[ 75%] Linking CXX shared module ../../../lib/mingw_osgdb_osgearth_splat_mask.dll

CMakeFiles/osgdb_osgearth_splat_mask.dir/objects.a(SplatMaskDriver.cpp.obj):SplatMaskDriver.cpp:(.text+0xf3d): undefined reference to `osgEarth::ProfileOptions::mergeConfig(osgEarth::Config const&)'

collect2.exe: error: ld returned 1 exit status

make[2]: *** [src/osgEarthDrivers/splat_mask/CMakeFiles/osgdb_osgearth_splat_mask.dir/build.make:120: lib/mingw_osgdb_osgearth_splat_mask.dll] Error 1

make[1]: *** [CMakeFiles/Makefile2:3189: src/osgEarthDrivers/splat_mask/CMakeFiles/osgdb_osgearth_splat_mask.dir/all] Error 2

Other plugins fail in a similar way.

gwaldron:
Why does this patch solve the problem?

jasonbeverage:
Merging this b/c it's a good fix if it solves your linker error and shouldn't effect anything else, but agree with Glenn that if you could explain what the actual issue is that would help us avoid it in the future.  Thanks!

filnet:
To be honest, I do not fully understand why linking fails. Could be a gcc 6.3.0 issue.

Using nm I could confirm that the symbol is missing from osgEarth library (which makes sense because of its "inline" nature).  But could not get why the splat_mask plugin fails to find it...

PR #908, opened 2017-04-12
Polytope mask stack no longer grows without bound in getWorldBounding…

emminizer:
…Polytope.


jasonbeverage:
Great fix Dan, thanks for tracking that down.

PR #909, opened 2017-04-13
GeoData: Support bounds that cross the antimeridian.

emminizer:


emminizer:
I have a TIFF file that cross the dateline.  Its data extents indicated a west of about -176 and east of +178, and returned false to crossesAntimeridian.  This change is an attempt to track down the root causes of this problem.  I did not detect any negative consequences to these changes but I am not as familiar with the potential consequences as you.  This is a possible solution, and possibly a good solution if there are no seen negative consequences.

jasonbeverage:
@emminizer these changes look great to me, thanks for tracking these down.  The change to transformExtentToMBR is definitely necessary.

PR #916, opened 2017-05-16
Fix broken download link for AlphaPixel

gb96:
Previous link was now a 404. Went to alphapixel.com and found current link to their downloads of prebuilt OSG and 3rd party dependencies


gwaldron:
Thanks gb!

PR #921, opened 2017-06-09
Windows GLCore/GL3 support and GL3 compile fixes

remoe:
This support Core GL3 compile on Windows. It is the same as in OSG:

http://forum.openscenegraph.org/viewtopic.php?t=13457
https://github.com/openscenegraph/OpenSceneGraph/blob/5d6322da2b84470fc0675b1c637246acb58802e9/CMakeModules/OsgMacroUtils.cmake#L214

I've testes it on Windows x64 with master branch and OSG 3.5.6 release.

About the GL3 fixes:

GL_DEPTH_BITS and GL_CLIP_PLANE0  is deprecated on GL3. And it doesn't build with core profile without this changes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -208,6 +208,11 @@ ENDIF(UNIX)
 
 
 IF(WIN32)
+  FIND_PACKAGE(GLCORE)
+  IF(GLCORE_FOUND)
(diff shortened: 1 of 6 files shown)


gwaldron:
At some point we'll need to figure out the clip plane thing, but that can wait. Thanks Remo!

PR #923, opened 2017-06-14
Topic gles3

tomhog:
Hi Glenn

So here are the changes to get gles3 working. Mostly changes to the shader merging, ensuring version number define is used everywhere and some work on static builds.

It also now generates the iOS example from CMake so should be a lot more user friendly.

I'm having issues now with tiles not paging in properly. It was working before but something seems to have crept in and I keep having issues with trianglefunctor type code crashing the app since osg added that new triange code. I just want to get this merged though as none of this has anything to do with that I it's getting a bit difficult to keep it up to date with the osgearth master.

Thanks
Tom


tomhog:
So the issue I was having is caused by osg and not related to osgearth. So this is working.

PR #927, opened 2017-06-23
Build error fix on Windows

emminizer:


gwaldron:
already fixed, thanks Dan ;)

PR #931, opened 2017-07-06
Add resample key to ResampleFilterOptions configuration

timoore:
Without "resample" as the value of the Config key, the proper driver
won't be found.

This fixes the osgearth_city demo.


gwaldron:
Thanks Tim!

PR #947, opened 2017-08-04
Update CMakeLists.txt

beachwalker:
Added missing featureeditor project.


gwaldron:
It's not missing; we removed it from CMakeLists on purpose.

PR #955, opened 2017-08-22
Fixes gdal-mutex crash-on-exit.

pgholm:
On gcc global cleanup local static variables are deleted before global static variables. To ensure that gdal_mutex is alive when registry destructs it has been moved into the class Registry.

### DIFF
--- a/src/osgEarth/Registry
+++ b/src/osgEarth/Registry
@@ -34,7 +34,7 @@
 #include <set>
 
 #define GDAL_SCOPED_LOCK \
-    OpenThreads::ScopedLock<OpenThreads::ReentrantMutex> _slock( osgEarth::getGDALMutex() )\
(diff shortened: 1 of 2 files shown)


jasonbeverage:
Thanks for looking into this!  I'm hesitant to take this change though as is without some further investigation.

We've had issues with applications crashing on exit related to the GDAL mutex before, and we've tried moving it around to various places that will fix one application but cause a crash in another.  It's all related to the static creation and destruction order of variables, so we need to look a bit further into the relationship between the various places in osgEarth that use the gdal mutex and ensure that everything is created and destroyed in the right order.

emminizer:
We've been experiencing sporadic issues on application close on Windows for the last 6-9 months or so.  This wasn't a problem before then.  I can test on my side to see if this helps our crashing bug if that's helpful.

PR #956, opened 2017-08-23
fix for OSGEARTH_SPLAT_EXPORT

shelltdf:
fix for OSGEARTH_SPLAT_EXPORT


gwaldron:
Thanks shelltdf - but why not just fix the one define at Export:45 instead of changing all the others?

shelltdf:
my English is not good
sorry for my carelessness
do you just want to modify "Export" file
cancel it or you fix it

gwaldron:
I can fix it. Thanks.

PR #958, opened 2017-08-25
Fixes assumption that the text object is the first child of Geode

pgholm:
Code will now look through all children and apply to the first one found.


gwaldron:
Thanks, merged!

PR #975, opened 2017-09-13
Cache wms, wfs, and tms metadata per service

johnlarring:
If a system has intermittent network issue (e.g. like at an IBC booth), you might not get some layer in an earth file at all if network connection breaks when metadata for layers are read (e.g. WFS Capabilities), event though all tiles are cached locally.

This commit is a hackish solution that caches WFS/WMS/TMS metacast data to disc when read successfully. If then at a later load of an earth file, the network connection is broken at the time when metadata is read, the cached metadata is used.

We would like to see these changes evolve into a propper feature, so that:
* metadata caching can be configured (Enable/disable, age-threshhold, etc.)
* solution is made platform indepedent and consistent with osgEarth data cache config.

### DIFF
--- a/src/osgEarthUtil/TMS.cpp
+++ b/src/osgEarthUtil/TMS.cpp
@@ -461,11 +461,34 @@ TileMapReaderWriter::read( const std::string& location, const osgDB::Options* op
     ReadResult r = URI(location).readString(options);
     if ( r.failed() )
     {
-        OE_DEBUG << LC << "Failed to read TMS tile map file from " << location
+	OE_DEBUG << LC << "Failed to read TMS tile map file from " << location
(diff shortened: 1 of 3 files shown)


gwaldron:
Thanks John. The proper approach here is to use the URI class when reading things like the WMS capabilities -- then they will go through the cache like other HTTP requests. I can work on that if you are interested.

johnlarring:
Hi Glenn, would be great if you could implement a solution for capability / metadata caching in the way you would like to have it integrated. We would like to see this feature in the next stable release. Until then, we'll just patch with my hack.

gwaldron:
John,
I've pushed a fix to TerrainLayer that was preventing the initialization data (like WMS Capabilities) from storing in the cache. Hopefully this will help.

PR #984, opened 2017-10-03
Don't set OPENSCENEGRAPH_VERSION in CMakeLists.txt

manisandro:
It is already set by DETECT_OSG_VERSION, and OPENSCENEGRAPH_{MAJOR,MINOR,PATCH}_VERSION are not defined anymore since b97eda03, resulting in OPENSCENEGRAPH_VERSION = ".."

Note that currently the logic to pick osgQt5 vs osgQt in https://github.com/gwaldron/osgearth/blob/master/CMakeModules/FindOSG.cmake#L122 is borken due to OPENSCENEGRAPH_VERSION being incorrectly set.


manisandro:
Superseded by c36847390c06cf70efdf12ab497b1c7c16fb8f0c

PR #991, opened 2017-10-16
Fix: OPENSCENEGRAPH_VERSION is null

KangLin:
Fix: OPENSCENEGRAPH_VERSION is null

Because set  OPENSCENEGRAPH_VERSION in OsgEarthMacroUtils.cmake


KangLin:
Because set  OPENSCENEGRAPH_VERSION in OsgEarthMacroUtils.cmake

PR #997, opened 2017-11-09
Allow a folder to be referrer

johnlarring:
This is very useful when embedding earth files into other files though ReaderWriter::writeNode(const osg::Node&, std::ostream&, const Options*). This allows us to have a fix setup root reference for all resources.

### DIFF
--- a/src/osgEarthDrivers/earth/EarthFileSerializer2.cpp
+++ b/src/osgEarthDrivers/earth/EarthFileSerializer2.cpp
@@ -27,6 +27,7 @@
 #include <osgDB/FileNameUtils>
 #include <stdio.h>
 #include <ctype.h>
+#include <sys/stat.h>
 
 using namespace osgEarth_osgearth;
 using namespace osgEarth;
@@ -181,7 +182,12 @@ namespace
         {
             _rewriteAbsolutePaths = false;
             _newReferrerAbsPath = osgDB::convertFileNameToUnixStyle( osgDB::getRealPath(referrer) );
-            _newReferrerFolder  = osgDB::getFilePath( osgDB::findDataFile(_newReferrerAbsPath) );
+	    // Check whether referrer is file or folder.
+	    struct stat sb;
+	    if (stat(_newReferrerAbsPath.c_str(), &sb) == 0 && S_ISDIR(sb.st_mode))
+		_newReferrerFolder = _newReferrerAbsPath;
+	    else
+		_newReferrerFolder  = osgDB::getFilePath( osgDB::findDataFile(_newReferrerAbsPath) );
         }
 
         /** Whether to make absolute paths into relative paths if possible */


gwaldron:
Can you explain this in a little more detail so I make sure we understand the use case?

johnlarring:
The usecase is where .earth file is streamed as an embedded element into a larger datastructure (project) that will be written to disc at a later stage. Hence, at the point where we need to serialize the MapNode, the referrer file does not yet exist. An one existing referrer file results file references within the MapNode layers to be written with absolute paths. This is why we want to be able to set a directory as referrer.

gwaldron:
John,
Unfortunately I had to revert this change because it broke serialization. The new code is removing forward slashes in some URLs. You can reproduce by running "osgconv readymap.earth out.earth" and see that the URLs are missing slashes. I don't have time atm to track down the core problem so I'll punt it back to you for now :)

johnlarring:
Hi Glenn,

No luck reproducing the issue. The TMS urls to readymap are serialized just fine. What exactly happened to URLs in you test. If it is the windows implementation POSIX "stat" that messes the URL up, then please try to following work-around. This pr and use-case are important to us :

```
diff --git a/src/osgEarthDrivers/earth/EarthFileSerializer2.cpp b/src/osgEarthDrivers/earth/EarthFileSerializer2.cpp
index 75ff516..4608349 100644
--- a/src/osgEarthDrivers/earth/EarthFileSerializer2.cpp
+++ b/src/osgEarthDrivers/earth/EarthFileSerializer2.cpp
@@ -182,9 +182,16 @@ namespace
         {
             _rewriteAbsolutePaths = false;
             _newReferrerAbsPath = osgDB::convertFileNameToUnixStyle( osgDB::getRealPath(referrer) );
-           // Check whether referrer is file or folder.
            struct stat sb;
+#if 1
+           // If referrer is local (i.e. starts with slash, not "<protocol>://"), then
+           // check whether referrer is file or folder.
+           if (!_newReferrerAbsPath.empty() && (_newReferrerAbsPath.substr(0,1) == "/")
+                   && stat(_newReferrerAbsPath.c_str(), &sb) == 0 && S_ISDIR(sb.st_mode))
+#else
+           // Check whether referrer is file or folder.
            if (stat(_newReferrerAbsPath.c_str(), &sb) == 0 && S_ISDIR(sb.st_mode))
+#endif
                _newReferrerFolder = _newReferrerAbsPath;
            else
                _newReferrerFolder  = osgDB::getFilePath( osgDB::findDataFile(_newReferrerAbsPath) );
```

gwaldron:
Run:
osgconv readymap.earth out.earth

The output contains the URL: "http:/readymap.org/readymap/tiles/1.0.0/7"

(missing the double-slash after the http as well as the trailing slash found in the original)

PR #1003, opened 2017-11-13
Fix bug in ElevationLayerVector::populateHeightFieldAndNormalMap

timoore:
commit 4b897935616a74503cf86e200a813fc24fc37410 optimized copying of
TileKey objects by using references, but these references could
clobber keys in the contenders array that will still be used. Scratch
space for actualKey is now used if necessary.

If the border code is re-enabled, then this bug will need to be fixed for
contenderKey as well.

The test case I used was osgearth_lights boston_projected.earth. The
bug prevents buildings from appearing.


jasonbeverage:
Thanks Tim, good find.

PR #1013, opened 2017-11-30
PixelAutoTransform: LOD Scale now only applies if changing scale, and…

emminizer:
… it resets back after cull traversal.


emminizer:
We were seeing a problem in orthographic overhead mode in the SIMDIS SDK with this code.  We need to dynamically adjust the LOD scale based on perceived eye position vs actual eye position in Ortho mode to fix the LODs on loaded entity models.  We got it working, but because PixelAutoTransform is a child, it was resetting our LOD scale back to 1.  I also saw an issue where PAT was changing the LOD scale not just for its children, but for everything else in the cull visitation after it.

Sorry for the short update, had test code in to ensure this was required, and I forgot to take it out before initial PR.

PR #1016, opened 2017-12-05
 use osgDB::readRef*File() in place of osgDB::read*File() 

filnet:


filnet:
Fails to compile with osg 3.4.0.

/home/travis/build/gwaldron/osgearth/src/osgEarth/Cache.cpp:171:95: error: conversion from osg::ref_ptr<osg::Object> to non-scalar type osgDB::ReaderWriter::ReadResult requested
         osgDB::ReaderWriter::ReadResult rr = osgDB::readRefObjectFile( driverExt, rwopt.get() );

Will look into it.

filnet:
Seems ok now.

PR #1018, opened 2017-12-05
DrapeableNode is now a MapNodeObserver.

emminizer:
This fixes two problems:
- DrapeableNode may not be in the NodePath of a MapNode, so its update traversal never finds an initial MapNode.  setMapNode() gives users a way to configure it properly externally.
- The Map Node may change at runtime and DrapeableNode would not catch the change.

### DIFF
--- a/src/osgEarth/DrapeableNode
+++ b/src/osgEarth/DrapeableNode
@@ -23,6 +23,7 @@
 #include <osgEarth/Common>
 #include <osgEarth/optional>
 #include <osgEarth/MapNode>
+#include <osgEarth/MapNodeObserver>
 #include <osg/Group>
 
 namespace osgEarth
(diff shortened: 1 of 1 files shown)


gwaldron:
Dan, the long-term plan was to get away from the MapNodeObserver pattern (which I think is messy) and eventually just require that AnnotationNodes (or any kind of Drapable/Clampable) be a descendant of MapNode. I feel like this makes sense since they depend on the map/terrain for their operation anyway. Thoughts?

emminizer:
How would you deal with change in the MapNode?  For example, when the MapNode changes in this case, the observer_ptr becomes invalidated, but the update traversal has already been turned off so you'll not catch new MapNodes.  You could keep update traversal on, but there's a performance cost to that and it results in unnecessary polling, since 99.99% of the time on update you won't have a map change.

Maybe it's messy, but it's also more cost efficient than the update traversal approach.

It would also more deeply complicate what we're currently during with the node layout in simVis::SceneManager.  I do not offhand know the ramifications of making everything a child of the map again.  I do know that we'd need to be more careful about displacing the old MapNode with the new MapNode, replacing children as needed.  Shouldn't be a big deal, but I just don't know offhand whether we'll have any graphics ramifications of this.  For example, we turn off depth writing in some cases on the terrain, but want it turned back on for entities.  We'll just need to make sure we cover that on our side.

gwaldron:
OK, so lots of issues there. I'll merge this for now and we can discuss it all at a later time. In a perfect world there would be no need to replace a MapNode at all ... rather just call setMap etc. and move on.

PR #1025, opened 2018-01-04
Adds optional name property in osgEarth::Symbology::StyleSheet

johnlarring:
Enables applications to keep track of what style sheet was applied to which feature model layers.


jasonbeverage:
Thanks John!

PR #1028, opened 2018-01-11
[WIP] Notify: add the possibility to set a notification handler

filnet:
and brings it up to par with OSG's Notify

this is basically a copy/paste of the current osg::Notify adapted to osgEarth.

tested against osg 3.5.10, not tested against osg 3.4.0.

### DIFF
--- a/src/osgEarth/Notify.cpp
+++ b/src/osgEarth/Notify.cpp
@@ -18,143 +18,225 @@
  */
 #include <osgEarth/Notify>
 
-using namespace osgEarth;
-
-/* -*-c++-*- OpenSceneGraph - Copyright (C) 1998-2006 Robert Osfield 
- *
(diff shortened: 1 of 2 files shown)


src/osgEarth/Notify.cpp
```diff
@@ -18,143 +18,225 @@
...
+
+osg::NotifyHandler* osgEarth::getNotifyHandler()
+{
+    NotifyStreamBuffer *buffer = static_cast<NotifyStreamBuffer *>(getNotifySingleton()._notifyStream.rdbuf());
+    return buffer ? buffer->getNotifyHandler() : 0;
+}
 
-    if (severity<=osgearth_g_NotifyLevel)
+
+#ifndef OSG_NOTIFY_DISABLED
```
filnet: need change that to OSGEARTH_NOTIFY_DISABLED

PR #1035, opened 2018-01-19
Used unused config URI::option_string to configure custom CURL headers

johnlarring:
The prime reason for sending custom curl headers through the unused URI::option_string is to enable configuration of oauth2 bearer token for sources protected by oauth2 authentication (Ref. ECMWF data at https://thredds.met.no/thredds/metno.html).

Example imagelayer accessing WMS restricted by oauth2:

```
    <image name="met-no-ecmwf-air-temperature" driver="wms">
        <url>https://thredds.met.no/thredds/wms/ecmwf/atmo/ec_atmo_0_1deg_20180118T120000Z_pl.nc</url>
	<option_string>Authorization: Bearer YOUR-TOKEN-STRING</option_string>
        <format>png</format>
	<version>1.3.0</version>
        <layers>air_temperature_pl</layers>
        <tile_size>256</tile_size>
        <srs>EPSG:4326</srs>
        <transparent>true</transparent>
        <times>2018-01-18T18:00:00.000Z</times>
        <cache_policy usage="no_cache"/>
    </image>
```
Passing URI::option_string as CURL custom header has been enabled for: TFS. WFS, TMS, WMS, and XYZ.

Note 1: Instead of using URI::option_string for passing CURL custom headers, it could be preferable to add URI::curl_header config, or even make it possible to add a list of custom CURL headers.

Note 2: CURLOPT_XOAUTH2_BEARER was not used, since it is not supported by current curl version on CentOS7 (curl-7.29.0)

Note 3: It would be possible to omit using osgDB::Options::pluginStringData if urls were passed as URI classes rather than std::string. However, this would be a much larger code change.

### DIFF
--- a/src/osgEarthDrivers/feature_tfs/FeatureSourceTFS.cpp
+++ b/src/osgEarthDrivers/feature_tfs/FeatureSourceTFS.cpp
@@ -81,10 +81,21 @@ public:
         // make a local copy of the read options.
         _readOptions = Registry::cloneOrCreateOptions(readOptions);
 
+	const URI tfsURI = _options.url().value();
+	if ( tfsURI.empty() )
+	{
+	    return Status::Error( Status::ConfigurationError, "Fail: driver requires a valid \"url\" property" );
+	}
+
+	// Add URI::option_string as plugin string data to be passed as custom header to CURL
+	// later in HTTPClient::doGet().
+	if (tfsURI.optionString().isSet())
+	    _readOptions->setPluginStringData("osgEarth::URI::optionString", tfsURI.optionString().get());
+
         FeatureProfile* fp = 0L;
 
         // Try to read the TFS metadata:
-        _layerValid = TFSReaderWriter::read(_options.url().get(), _readOptions.get(), _layer);
+	_layerValid = TFSReaderWriter::read(tfsURI, _readOptions.get(), _layer);
 
         if (_layerValid)
         {
--- a/src/osgEarthDrivers/xyz/ReaderWriterXYZ.cpp
+++ b/src/osgEarthDrivers/xyz/ReaderWriterXYZ.cpp
@@ -60,11 +60,16 @@ public:
       {
           _dbOptions = Registry::instance()->cloneOrCreateOptions(dbOptions);        
 
-          URI xyzURI = _options.url().value();
+	  const URI xyzURI = _options.url().value();
           if ( xyzURI.empty() )
           {
               return Status::Error( Status::ConfigurationError, "Fail: driver requires a valid \"url\" property" );
           }
+
+	  // Add URI::option_string as plugin string data to be passed as custom header to CURL
+	  // later in HTTPClient::doGet().
+	  if (xyzURI.optionString().isSet())
+	      _dbOptions->setPluginStringData("osgEarth::URI::optionString", xyzURI.optionString().get());
 
           // driver requires a profile.
           if ( !getProfile() )
--- a/src/osgEarth/HTTPClient.cpp
+++ b/src/osgEarth/HTTPClient.cpp
@@ -1136,8 +1136,14 @@ HTTPClient::doGet(const HTTPRequest&    request,
         }
     }
 
+    // Interpret osgEarth::URI::optionString as Curl custom header
+    std::string custom_header = options->getPluginStringData("osgEarth::URI::optionString");
+    if (!custom_header.empty())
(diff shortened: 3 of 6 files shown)


src/osgEarthDrivers/feature_tfs/FeatureSourceTFS.cpp
```diff
@@ -81,10 +81,21 @@ class TFSFeatureSource : public FeatureSource
...
+	// Add URI::option_string as plugin string data to be passed as custom header to CURL
+	// later in HTTPClient::doGet().
+	if (tfsURI.optionString().isSet())
+	    _readOptions->setPluginStringData("osgEarth::URI::optionString", tfsURI.optionString().get());
+
         FeatureProfile* fp = 0L;
 
         // Try to read the TFS metadata:
-        _layerValid = TFSReaderWriter::read(_options.url().get(), _readOptions.get(), _layer);
+	_layerValid = TFSReaderWriter::read(tfsURI, _readOptions.get(), _layer);
```
filnet: indentation got broken (note that the same issue appears multiple times in this PR)

src/osgEarthDrivers/feature_tfs/FeatureSourceTFS.cpp
```diff
@@ -81,10 +81,21 @@ class TFSFeatureSource : public FeatureSource
...
 
+	const URI tfsURI = _options.url().value();
+	if ( tfsURI.empty() )
+	{
+	    return Status::Error( Status::ConfigurationError, "Fail: driver requires a valid \"url\" property" );
+	}
+
+	// Add URI::option_string as plugin string data to be passed as custom header to CURL
+	// later in HTTPClient::doGet().
+	if (tfsURI.optionString().isSet())
```
filnet: missing braces

src/osgEarthDrivers/xyz/ReaderWriterXYZ.cpp
```diff
@@ -60,12 +60,17 @@ class XYZSource : public TileSource
...
           if ( xyzURI.empty() )
           {
               return Status::Error( Status::ConfigurationError, "Fail: driver requires a valid \"url\" property" );
           }
 
+	  // Add URI::option_string as plugin string data to be passed as custom header to CURL
+	  // later in HTTPClient::doGet().
+	  if (xyzURI.optionString().isSet())
+	      _dbOptions->setPluginStringData("osgEarth::URI::optionString", xyzURI.optionString().get());
+
```
filnet: indentation seems off

gwaldron:
John, the option_string property isn't unused. It is specifically for passing plugin-specific information to a OSG loader (jpeg, png, etc.) via the osgDB::Options::setOptionString() call. Because of this, I tend to agree with your "Note 1" in that custom HTTP headers will likely require a property separate from option_string.  But it's up for discussion :)

johnlarring:
@filnet sorry about the indentation issue, but my IDE is configured for a very different code style than osg/osgEarth, so it's hard to get them right. I'll try to get them fixed. 

@gwaldron Thanks for clarification about option_string. I could not find doc or usage in code, so I just grabbed it. We need a speedy resolution for oauth authentication, so if I change the pull-request to create a new "optional<std::string> _url_header" to osgEarth::URI w/config handling, will you be happy with that?

johnlarring:
Hi @gwaldron and @filnet , hope the updated pull-request will make it :-) Custom http header is now passed through URL::http_header instead of URL::option_string.

Also: added braces and reviewed indentation.

filnet:
Hi John,
You are indenting with a combination of tabs and spaces. Does not play well in all cases.
osgEarth is indented with spaces only.

Not sure it is a big deal...

filnet:
But you'll need to rework the commit history. You'll probably want to squash the two commits into one and change the commit log to reflect what you did in the end.

gwaldron:
John,
We discussed this internally. I think the approach we want to take is to add a set of custom "fields" to the URIContext object. These would be name/value pairs that the HTTPClient could then interpret as headers. Being in the URIContext would ensure that they get propagated nicely -- since it's likely that if you want the custom headers for one operation, relative-URI operations will want the same set of headers. 

I think it makes the most sense to incorporate this post-2.9 at this point.

johnlarring:
That's ok. We'll just build osgearth patched with https://github.com/gwaldron/osgearth/pull/1037 until custom fields in the URIContext object is implemented from you end.

PR #1037, opened 2018-01-22
Added new optional http_header config to osgEarth::URI

johnlarring:
The new URL::http_header config is used to pass custom http header to Curl in osgEarth::HTTPClient. 

The prime reason for adding the URI::http_header is to enable configuration of oauth2 bearer token for sources protected by oauth2 authentication (Ref. ECMWF data at https://thredds.met.no/thredds/metno.html).

Example imagelayer accessing WMS restricted by oauth2:
```
   <image name="met-no-ecmwf-air-temperature" driver="wms">
      <url>https://thredds.met.no/thredds/wms/ecmwf/atmo/ec_atmo_0_1deg_20180118T120000Z_pl.nc</url>
      <http_header>Authorization: Bearer YOUR-TOKEN-STRING</http_header>
      <format>png</format>
      <version>1.3.0</version>
      <layers>air_temperature_pl</layers>
      <tile_size>256</tile_size>
      <srs>EPSG:4326</srs>
      <transparent>true</transparent>
      <times>2018-01-18T18:00:00.000Z</times>
      <cache_policy usage="no_cache"/>
   </image>
```

Passing URI::http_header to CURL has been enabled for: TFS. WFS, TMS, WMS, and XYZ.

Note 1: CURLOPT_XOAUTH2_BEARER was not used, since it is not supported by current curl version on CentOS7 (curl-7.29.0)

Note 2: It would be possible to omit using osgDB::Options::pluginStringData if urls were passed as URI classes rather than std::string. However, this would be a much larger code change.

### DIFF
--- a/src/osgEarth/HTTPClient.cpp
+++ b/src/osgEarth/HTTPClient.cpp
@@ -1136,8 +1136,16 @@ HTTPClient::doGet(const HTTPRequest&    request,
         }
     }
 
+    // Interpret osgEarth::URI::optionString as Curl custom header
+    std::string custom_header = options->getPluginStringData("osgEarth::URI::httpHeader");
+    if (!custom_header.empty())
+    {
+	headers = curl_slist_append(headers, custom_header.c_str());
+    }
+
     // Disable the default Pragma: no-cache that curl adds by default.
     headers = curl_slist_append(headers, "Pragma: ");
+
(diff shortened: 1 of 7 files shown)


gwaldron:
Believe this was resolved in another way - close?

gwaldron:
Resolved by 2fb0bac2c5b2e56136008bf057c14cc79e48efac.

PR #1041, opened 2018-01-30
rex: fix compilation error with gcc 7.3.0

filnet:


filnet:
A similar fix just landed in master. Closing.

PR #1044, opened 2018-02-02
EarthManipulator: single click actions now correctly break tether

mk401:
Encountered when trying to break camera tether with a mouse click.

`bindMouseClick(EarthManipulator::ACTION_GOTO, osgGA::GUIEventAdapter::LEFT_MOUSE_BUTTON, osgGA::GUIEventAdapter::MODKEY_CTRL, goToOpt);`

Without this change, the bound mouse click doesn't correctly untether the camera.

### DIFF
--- a/src/osgEarthUtil/EarthManipulator.cpp
+++ b/src/osgEarthUtil/EarthManipulator.cpp
@@ -1720,6 +1720,13 @@ EarthManipulator::handle(const osgGA::GUIEventAdapter& ea, osgGA::GUIActionAdapt
         const ActionTypeVector& atv = _settings->getBreakTetherActions();
         if ( atv.size() > 0 )
         {
+            int eventType = ea.getEventType();
+            // use custom EVENT_MOUSE_CLICK enum value if left mouse button click
+            // is detected, because it is used in Settings::bindMouseClick()'s InputSpec.
+            if ( eventType == osgGA::GUIEventAdapter::LEFT_MOUSE_BUTTON )
+            {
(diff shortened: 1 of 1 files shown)


gwaldron:
This won't work -- eventType is never used, and LEFT_MOUSE_BUTTON is not a valid return value for ea.getEventType(). 

But I get what you are trying to do. Let me work up a fix.

mk401:
You're right, it won't work. I intended to put eventType in the call to settings_->getAction() a few lines down. That's what I get for not double checking my pull request was complete.

gwaldron:
Please try this on your end and let me know if it works:

```
        if ( atv.size() > 0 )
        {
            EventType eventType = (EventType)ea.getEventType();
            if (isMouseClick(&ea))
                eventType = EVENT_MOUSE_CLICK;

            const Action& action = _settings->getAction( eventType, ea.getButtonMask(), ea.getModKeyMask() );
            if ( std::find(atv.begin(), atv.end(), action._type) != atv.end() )
            {
                clearViewpoint();
            }
        }

```

mk401:
No, that doesn't solve it. That change also breaks our double-click-to-untether.

gwaldron:
Try:
`if (eventType == osgGA::GUIEventAdapter::RELEASE && isMouseClick(&ea) )`

mk401:
No dice. Double click works with that change, but not single click.

gwaldron:
Ok. I broke down and wrote a test case myself. Pushed a fix.

mk401:
Yep, that works great. Thank you!

PR #1048, opened 2018-02-06
cache_rocksdb: fix linking issues on msys2/mingw with latest rocksdb

filnet:
Compile cache_rocksdb with latest rocksdb had two issues:
- cmake could not find librocksdb as it was searching for rocksdblib
- linking cache_rocksdb failed because libraries needed by rocksdb were listed before rocksdb

This PR attempts to address both issues.
With the proposed changes osgearth and cache_rocksdb compile fine.


remoe:
I've done the cmake update also :)

https://github.com/gwaldron/osgearth/pull/1042

filnet:
@remoe Yep, saw that. You'll go first and I'll fix the conflict on my side.

PR #1068, opened 2018-03-12
Map: addLayer() now puts the new layer in a ref_ptr to prevent accide…

emminizer:
…ntal deletion during callback processing.


emminizer:
This fixes a crash bug when Elevation Layers in a .earth file are set with visible=false.  Without this, the first pass of map callbacks ends up putting the layer in a ref_ptr that is then discarded, which deletes the layer mid-processing of callbacks, causing a segfault.

PR #1077, opened 2018-03-23
Fix build with libgeos using MinGW

tehnick:
Fixed linking of osgEarthSymbology with geos when build as shared library using MinGW.


filnet:
Hi,

I am a maintainer of the MSYS2/MinGW osgEarth package.
osgEarth builds fine there without GEOS_INLINE defined so I was wondering what issue you had and what that define does and how it fixes the issue you are facing ?

tehnick:
Hi @filnet ,

> I am a maintainer of the MSYS2/MinGW osgEarth package.
> osgEarth builds fine there without GEOS_INLINE defined

Could you show your build rules?

> I was wondering what issue you had

```
[ 26%] Built target osgEarth
[ 26%] Linking CXX shared library ../../lib/libosgEarthSymbology.dll
CMakeFiles/osgEarthSymbology.dir/objects.a(GEOS.cpp.obj):GEOS.cpp:(.text+0xda): undefined reference to `geos::geom::PrecisionModel::~PrecisionModel()'
CMakeFiles/osgEarthSymbology.dir/objects.a(GEOS.cpp.obj):GEOS.cpp:(.text+0x248): undefined reference to `geos::geom::Coordinate::Coordinate(double, double, double)'
CMakeFiles/osgEarthSymbology.dir/objects.a(GEOS.cpp.obj):GEOS.cpp:(.text+0x4ce): undefined reference to `geos::geom::GeometryFactory::getCoordinateSequenceFactory() const'
CMakeFiles/osgEarthSymbology.dir/objects.a(GEOS.cpp.obj):GEOS.cpp:(.text+0x55e): undefined reference to `geos::geom::GeometryFactory::getCoordinateSequenceFactory() const'
CMakeFiles/osgEarthSymbology.dir/objects.a(GEOS.cpp.obj):GEOS.cpp:(.text+0x6ab): undefined reference to `geos::geom::GeometryFactory::getCoordinateSequenceFactory() const'
CMakeFiles/osgEarthSymbology.dir/objects.a(GEOS.cpp.obj):GEOS.cpp:(.text+0x714): undefined reference to `geos::geom::GeometryFactory::getCoordinateSequenceFactory() const'
collect2: error: ld returned 1 exit status
make[2]: *** [src/osgEarthSymbology/CMakeFiles/osgEarthSymbology.dir/build.make:1208: lib/libosgEarthSymbology.dll] Error 1
make[1]: *** [CMakeFiles/Makefile2:331: src/osgEarthSymbology/CMakeFiles/osgEarthSymbology.dir/all] Error 2
make: *** [Makefile:152: all] Error 2
```

> what that define does

https://github.com/OSGeo/geos/blob/master/src/inlines.cpp

> and how it fixes the issue you are facing ?

It fixes linking with shared GEOS libraries.

> I am asking because the proposed change will affect the MSYS2/MINGW build too...

Ok, let's check it.

> PS : seems you where able to set GEOS_INLINE on your side (i.e. mxe) so I am wondering if this PR is still relevant ?

Usage of `-DCMAKE_CXX_FLAGS='-DGEOS_INLINE=1'` (as you see in https://github.com/mxe/mxe/commit/a118ee0271f9bd4c6f0a0d42b981e7d12213799a) is not a solution of problem, but just another way to workaround it. We will use it if necessary. But providing patches for upstream is more preferable way.

filnet:
This is how osgEarth is built by MSYS2/MinGW:

```
      ${MINGW_PREFIX}/bin/cmake.exe \
        -G"MSYS Makefiles" \
        -DCMAKE_INSTALL_PREFIX=${MINGW_PREFIX} \
        -DCMAKE_BUILD_TYPE=${builddir%-${MINGW_CHOST}} \
        -DOSG_DIR=${MINGW_PREFIX} \
        -DOSGEARTH_QT_BUILD=ON \
        -DOSGEARTH_QT_BUILD_LEGACY_WIDGETS=ON \
        ${_opts} \
        ../${_realname}-${_pkgver}
```
DYNAMIC_OSGEARTH is not set so will default to ON.

Currently it builds osgEarth 2.9 with GEOS 3.6.1.

filnet:
Could the issue come from how GEOS was built by mxe where some symbols were not exported ?

tehnick:
> Could the issue come from how GEOS was built by mxe where some symbols were not exported ?

In MXE the default configuration options are used.

> This is how GEOS is built by MSYS2/MinGW:
> ...
> ```    -DGEOS_ENABLE_INLINE=OFF \```
> ...

This is why you do not have a linking error. Why have you disabled `GEOS_ENABLE_INLINE`?

filnet:
>

filnet:
> Why have you disabled GEOS_ENABLE_INLINE?

I was not the one who did that but I guess it was done to avoid the issue you faced.
By building GEOS with GEOS_ENABLE_INLINE to ON you then force all other projects to enable it on their side when using GEOS.
The person who contributed GEOS to MSYS2/MinGW probably decided to not put that constraint on GEOS users (with a minor performance issue as a side effect).

tehnick:
@tonytheodore Could you add your comment here?

We may add `--disable-inline` configuration option to `geos` package in MXE to get the same result.

tonytheodore:
Ideally, the `geos-config` script should be modified to output the value it was configured with, for some reason it was [removed a long time ago](https://github.com/OSGeo/geos/commit/a2bdefc363f395802062afce2f5db78feb0dd1ef#diff-32c8c942502b5590247861e9af5dc980). Without that, disabling it in `geos` package seems reasonable.

tehnick:
> Without that, disabling it in `geos` package seems reasonable.

Ok, let's go this way.

PR #1079, opened 2018-03-25
Fixed bug of explicitly loading osgEarthUtil library in the ReaderWri…

Vic-Min:
…terOsgEarth.cpp file

In Debug configuration explicitly loaded Release version of osgEarthUtil library.

To reproduce the bug, run the command:

```
$ OSGEARTH_NOTIFY_LEVEL=DEBUG osgearth_viewerd simple.earth
[osgEarth]  [Earth Plugin] Forced load: libosgEarthUtil.so
…

```
Reason: incorrectly detected Debug configuration in the src/osgEarthDrivers/earth/ReaderWriterOsgEarth.cpp file. For this purpose, MS Visual Studio specific macro `_DEBUG` and Qt specific macro `QT_DEBUG` are used.
Also, RelWithDebInfo and MinSizeRel configuration are not handled.

In addition, in the src\osgEarthDrivers\fastdxt\CMakeLists.txt file defined `OSGEARTH_DEBUG_POSTFIX` and `OSGEARTH_RELEASE_POSTFIX` macro, but never used. They are removed to prevent misuse with RelWithDebInfo and MinSizeRel configuration.


jasonbeverage:
Thanks for tracking this down!

PR #1081, opened 2018-04-02
Replace several deprecated osg::Geometry::BIND_ with equivalent non-d…

emminizer:
…eprecated osg::Array::BIND_ enums.

### DIFF
--- a/src/osgEarth/ObjectIndex.cpp
+++ b/src/osgEarth/ObjectIndex.cpp
@@ -159,10 +159,9 @@ ObjectIndex::tagDrawable(osg::Drawable* drawable, ObjectID id) const
         return;
 
     // add a new integer attributer to store the feautre ID per vertex.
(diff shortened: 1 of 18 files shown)


emminizer:
I was having problems with the GL CORE 3.3 build of osgEarth and the SIMDIS SDK with some basic examples.  I tracked down some crashes and bad behavior in SDK to this type of code.

In particular, OSG does not seem to like the deprecated setNormalArrayBinding() call any more in 3.6.0-rc2+.  During runtime if in GL CORE without FFP, you see an error about invalid operations in GL.  This patch fixes a few of those errors.  I updated the color and vertex binding too to be similar.

I did not hit every location that was used.  There is some mesh code that still uses the osg::Geometry enums and I'm hesitant to change that because I'm not familiar with it at all, and it doesn't follow the patterns seen in this commit.

gwaldron:
I will take care of the remaining ones -- thanks.

PR #1084, opened 2018-04-04
GL_LIGHTING mode no longer set if FFP not available.  Now consistentl…

emminizer:
…y setting OE_LIGHTING_DEFINE.


emminizer:
I had attempted to do this automatically in ShaderGenerator but couldn't get it to consistently remove all the GL_LIGHTING flags that osgEarth was using/generating in GL3/No-FFP mode.

After this patch, I am no longer getting any lighting-related invalid enumeration errors in GL 3.3.

emminizer:
After some more research, it might be the case that we can drop this and replace with a single call to osg::State's setModeValidity(GL_LIGHTING, false).  This would need to be done for every graphics context.  I haven't tested it out yet but it might be a cleaner path forward.

gwaldron:
Apparently that setting's been there for 12 years .. first I've heard of it :)

gwaldron:
Dan, I took a different approach to the same issue. Closing this PR. Thanks.

emminizer:
No problem, I understand and agree with the approach I see you're taking.  Sounds good.

PR #1087, opened 2018-04-04
Simple Sky: Stars draw in GL Core in OSG 3.6.0.  Fixed periodic stars…

emminizer:
… flashing bug by removing the near/far ratio on stars camera.


emminizer:
I tested this in both 3.4.1 (against GL 4.5 with compatibility mode) and in 3.6.0-rc4 (against GL 3.3 with core mode).

This does backout part of the fixes from https://github.com/gwaldron/osgearth/commit/4e040c5b but this should be better supported, at least in OSG 3.4+ due to variable locations on the color array.  In my 3.6 testing the color attribs were at location 2, not 3.  OSG's replacement scheme in Program seems to take care of this.

The near/far clip plane removal fixes a flickering seen on some video cards, presumably due to a float precision issue with the clip plane.

PR #1091, opened 2018-04-09
Simple Sky: Fix potential NULL access.

emminizer:
Encountered a segfault where image might be NULL if the moon image is not found, leading to a crash in GL3 mode.  Mea culpa.


remoe:
I had now the same crash :) Thanks.

Issue #1098, opened 2018-04-10
osgearth_pick on MacOSX: Image layer not visible 


remoe:
After run `./osgearth_pick tests/feature_country_boundaries.earth` the image layer is not visible:

<img width="336" alt="screen shot 2018-04-10 at 12 43 33" src="https://user-images.githubusercontent.com/804941/38556048-97eed1ae-3cc8-11e8-9738-b8528450a588.png">

But it works if one run it with `osgearth_viewer`. Tested with `rex` and `mp`.

OSG: 3.6.0
osgearth: cc63ab23dbc97353b9cf46f1e93477278fb26076

emminizer:
For what it's worth, I just tested on Windows with GL Core profile and current master and I can see the earth.  So like your PR does, I'm guessing it's an apple-specific issue.

remoe:
Now, i can reproduce the issue with an OSG Sample:

https://github.com/remoe/osgsample-2

This works on Windows without any errors. On MacOSX it throws:

```ERROR: Active attribute aliasing. Slot 3 unavailable for 'osg_MultiTexCoord0' from BindAttributeLocation request.```

When one disable this:

https://github.com/remoe/osgsample-2/blob/master/main.cpp#L205

with "#if 0", Then it doesn't throw an error on MacOSX.

Tested with OSG 3.6.2 / GLCore.

remoe:
The main issue is the following: OSG bind both vertex attribute arrays to the same index:

Linking osg::Program "osgEarth.engine_mp.TerrainNode" id=13 contextID=0
**Program's vertex attrib binding 4, oe_index_objectid_attr**
Program's vertex attrib binding 6, oe_terrain_attr
Program's vertex attrib binding 7, oe_terrain_attr2
State's vertex attrib binding 2, osg_Color
State's vertex attrib binding 12, osg_FogCoord
State's vertex attrib binding 3, osg_MultiTexCoord0
**State's vertex attrib binding 4, osg_MultiTexCoord1**
State's vertex attrib binding 5, osg_MultiTexCoord2
State's vertex attrib binding 6, osg_MultiTexCoord3
State's vertex attrib binding 7, osg_MultiTexCoord4
State's vertex attrib binding 8, osg_MultiTexCoord5
State's vertex attrib binding 9, osg_MultiTexCoord6
State's vertex attrib binding 10, osg_MultiTexCoord7
State's vertex attrib binding 1, osg_Normal
[...]

PR #1099, opened 2018-04-10
MacOSX: fix for #1098

remoe:


jasonbeverage:
Hey Remo!

I believe you that this fixes the issue, but I want to understand why to see if we can do a more general fix.  The default attribute location for the ObjectIndex is SECONDARY_COLORS (which is attribute location 4).  Changing it to 8 is changing it to TEXTURE_COORDS_0, which doesn't seem like what you would want to do.  I believe the enum locations are setup to map to nvidia's driver attribute locations, but I'm sure that is different on OSX.  

Do you have a feel for what the core issue is here?  Is using attribute location 4 (SECONDARY_COLORS) stomping on the something in the osx driver that isn't an issue on windows?  What about using attribute 5 (FOG_COORDS), would that work instead?

remoe:
It's not a correct fix. I think something is wrong with setUseVertexAttributeAliasing in a GLCore environment. I see errors like on MacOSX:

ERROR: Active attribute aliasing. Slot 4 unavailable for 'oe_index_objectid_attr' from BindAttributeLocation request.

But i don't see this on Windows GLCore (OSG 3.6.2). setUseVertexAttributeAliasing  force to bind all this default attributes:

https://github.com/openscenegraph/OpenSceneGraph/blob/6cd00edd6c604dcb9156bf657560823cfdea07ec/src/osg/Program.cpp#L780

I think this is wrong. It should only bind used attributes.

PR #1100, opened 2018-04-10
Break long import_defines() call in a REX Engine shader into two pragmas

RALaFrance222:
To work around bug in old Intel Graphics drivers


gwaldron:
How on earth did you figure this out?

RALaFrance222:
We used API Trace to figure out that OSGEarth was encountering an issue during shader compilation, and then it was just a matter of pasting lines from the misbehaving shader into the compiler one at a time

PR #1101, opened 2018-04-10
Triton: GL3 fixes to get Ocean drawing with GL Core profile without e…

emminizer:
…rrors.

### DIFF
--- a/src/osgEarthTriton/TritonContext.cpp
+++ b/src/osgEarthTriton/TritonContext.cpp
@@ -126,12 +126,14 @@ TritonContext::initialize(osg::RenderInfo& renderInfo)
 
             float openGLVersion = osg::getGLVersionNumber();
             enum ::Triton::Renderer tritonOpenGlVersion = ::Triton::OPENGL_2_0;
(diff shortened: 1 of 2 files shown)


emminizer:
The openGLVersion check applies the correct Triton version.  However this is now protected against change from an ifndef on `OSG_GL_FIXED_FUNCTION_AVAILABLE`.  This is because out of the box, Triton 3.91 will fail with anything but OPENGL_2_0 due to an error in VAO handling.

The WATER_DECALS issue was a problem both in 3.91 and 4.01 in core profile.  It is not a problem in compatibility profile for 3.91, but based on my reading of the code there is typically not a shader installed here anyways.

The mode validity prevents OSG from setting `GL_LIGHTING` on the next state change.  This should only be needed in Core profile mode, as the flag is deprecated in 3.2

After this series of changes, I now have Triton drawing an ocean in examples in Linux g++ 4.4 and Windows MSVC 2015 x64 against 3.91.  I tested Linux in 4.01 but decided against upgrading for now, so I didn't get around to testing Windows on that side.

PR #1114, opened 2018-04-30
add description to DataExtent

aroth-fastprotect:
adds an optional string property "description" to DataExtent. This allows to add a human readable description to each data extent in tms.xml, since the coordinates are quite confusion sometimes.

### DIFF
--- a/src/osgEarth/GeoData
+++ b/src/osgEarth/GeoData
@@ -548,8 +548,11 @@ namespace osgEarth
     {
     public:
         DataExtent(const GeoExtent& extent);
+        DataExtent(const GeoExtent& extent, const std::string &description);
         DataExtent(const GeoExtent& extent, unsigned minLevel );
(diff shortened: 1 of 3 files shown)


jasonbeverage:
Hi Andreas!  

What is the use case for this?  The only project that actually produces a tms with a DataExtents array is our ReadyMap server and it just serves as a hint to osgEarth not to make tile requests where no data will ever exist.  So it's just a performance booster at the end of the day.

aroth-fastprotect:
Hi,

we are using a TMS repository for our production machines too. The actual data included in the TMS depends on the project and use cases. When you look at a TMS repository is difficult to see which areas are included in the TMS. Checking the tms.xml is an easy way, but only if you have the description (or if you understand the raw GPS bounds ;-) ).

jasonbeverage:
Sounds good to me.  I wouldn't rely too much on the hacked up TMS spec, but it's been working well for us for a long time :)

PR #1115, opened 2018-04-30
allow to get/reset proxy settings

aroth-fastprotect:
this allows to get the current proxy settings and allow to reset the proxy settings to not use a proxy at all


jasonbeverage:
Thanks Andreas!

PR #1119, opened 2018-05-07
Added OSG's Gen include dir to the include path.

DigitalInBlue:
This fixed a problem in VS2017 with my install not finding ```osg/Config```.


jasonbeverage:
Hey John are you getting this issue if you build against the result of the INSTALL target from OpenSceneGraph or are you trying to build against your build directory?  The results in the OSG_GEN_INCLUDE_DIR should be copied over to the final output path when you run INSTALL.

PR #1121, opened 2018-05-14
ScreenSpaceLayout: Update to internal variable name to match OSG 3.6.…

emminizer:
…1-rc4


emminizer:
See: https://github.com/openscenegraph/OpenSceneGraph/commit/510a231e145d0
Build errors against current OpenSceneGraph-3.6 branch without this change.  Tested by running "osgearth_tracks readymap.earth"

PR #1123, opened 2018-05-17
The vcproj version of leveldb requires Shlwapi

plevy:
The vcproj version of leveldb requires Shlwapi to resolve PathFileExistsW.  I am not sure if this is specific to the vcproj leveldb or VS2017 or both


filnet:
This PR duplicates https://github.com/gwaldron/osgearth/pull/1123.

Note that the other PR is better targeted (MSVC) while this one will probably break MINGW builds.

PR #1124, opened 2018-05-19
Added missing library for leveldb under Visual Studio

DigitalInBlue:
I had to add this library to leveldb to support its use of PathFileExists:

https://msdn.microsoft.com/en-us/library/windows/desktop/bb773584(v=vs.85).aspx

I assumed this was a windows-only bug, so I wrapped it in the CMakeLists.txt.  I am not sure your standard for how to handle this, so I took a stab at it.


filnet:
This PR duplicates #1123.

Note that the other PR is better targeted (MSVC) while this one will probably break MINGW builds.

jasonbeverage:
I've merged #1123

PR #1138, opened 2018-06-13
fast_map: Replaced vector with list to avoid data invalidation in thr…

emminizer:
…eaded use of PerObjectFastMap in HorizonClipPlane::operator().

### DIFF
--- a/src/osgEarth/Containers
+++ b/src/osgEarth/Containers
@@ -80,7 +80,7 @@ namespace osgEarth
         iterator begin() { return _data.begin(); }
         iterator end() { return _data.end(); }
         bool empty() const { return _data.empty(); }
-        void clear() { _data.clear(); }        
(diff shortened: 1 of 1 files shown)


emminizer:
osgearth_viewer was consistently crashing for me.  The cause was two threads grabbing data from the PerObjectFastMap in the same HorizonClipPlane.  This was a problem because when the second thread grabbed an entry, it caused the std::vector in fast_map to increment.  That invalidated iterators and references due to a resize in std::vector.  That caused the 'd' reference to become invalid in the first thread, leading to a crash once the first thread got control again.

This was mitigated by replacing the vector with a std::list, which will not invalidate the iterator / data values on insertion.  The primary downsides to this simple fix are that size() is now potentially O(n) (primarily on gcc on older c++ standards), and iteration is possibly slightly slower.

PR #1146, opened 2018-06-27
Fixed an issue where setting color for the first vertex in a LineDraw…

crmcnulty:
…able in GL_LINE_LOOP mode would not set the color for use in both directions.


crmcnulty:
image

Example of the issue

Issue #1147, opened 2018-06-27
Empty earth file crashes on some windows machines (mp engine)


remoe:
osgearth_viewer crashes on some windows machine with empty earth file:

```
set OSGEARTH_TERRAIN_ENGINE=mp
osgearth_viewer test.earth
```

test.earth
```xml
<map name="test" type="geocentric">
</map> 
```

the crash is undefined in the gpu driver. But this should not happen.

```
DrvPresentBuffers + 0x917184
DrvPresentBuffers + 0x1600706
DrvPresentBuffers + 0x1602251
DrvPresentBuffers + 0x1543486
osg::DrawElementsUShort::draw + 0x342
osg::Drawable::draw + 0x350
osgUtil::RenderLeaf::render + 0x197
osgUtil::RenderBin::drawImplementation + 0x306
osgUtil::RenderBin::draw + 0x110
osgUtil::RenderBin::drawImplementation + 0x435
osgUtil::RenderStage::drawImplementation + 0x466
osgUtil::RenderBin::draw + 0x110
osgUtil::RenderStage::drawInner + 0x228
osgUtil::RenderStage::draw + 0x1310
osgUtil::SceneView::draw + 0x3627
osgViewer::Renderer::draw + 0x848
```

The system has the following configuration:

```
osgEarth Library 2.10.0 DEVELOPMENT 

[osgEarth]  [Registry] Terrain engine set from environment: mp
[osgEarth]  [Capabilities] osgEarth Version: 2.10.0 DEVELOPMENT 
[osgEarth]  [Capabilities] Detected hardware capabilities:
[osgEarth]  [Capabilities]   Vendor = NVIDIA Corporation
[osgEarth]  [Capabilities]   Renderer = Quadro K420/PCIe/SSE2
[osgEarth]  [Capabilities]   Version = 3.3.0 NVIDIA 353.30
[osgEarth]  [Capabilities]   Core Profile = yes
[osgEarth]  [Capabilities]   Max GPU texture units = 32
[osgEarth]  [Capabilities]   Max GPU texture coord indices = 8
[osgEarth]  [Capabilities]   Max GPU attributes = 16
[osgEarth]  [Capabilities]   Max texture size = 16384
[osgEarth]  [Capabilities]   Max lights = 1
[osgEarth]  [Capabilities]   GLSL = yes
[osgEarth]  [Capabilities]   GLSL Version = 330
[osgEarth]  [Capabilities]   Texture arrays = yes
[...]

PR #1150, opened 2018-07-02
Fix for #1147

remoe:
I hope it's a "correct" fix :)


jasonbeverage:
Thanks for digging into to this Remo!

PR #1158, opened 2018-07-13
GL3 fixes for osg::Point spam for point-based Features.

emminizer:
Tested on OSG 3.4 with compatibility profile GL support and everything still works with point features.  Tested on OSG 3.6 with core profile GL support and although point size is still not respected, at least there is no more continuous error spam from unsupported osg::Point or GL_POINT_SMOOTH (a mode that is supported by osg::Point)


emminizer:
Though still marked as "Passed", this now fails on GL3 Core Profile OSG because `osg/Point` got moved into an `ifdef` in https://github.com/gwaldron/osgearth/commit/09d225315

I'm waiting to update the diff based on the outcome of `PointDrawable` changes currently underway, which might make this PR OBE.

PR #1159, opened 2018-07-16
Triton: TritonLayer now accepts Callback parameter.  osgearth_triton …

emminizer:
…app updated to use TritonLayer.


emminizer:
Note that setOpacity() on the layer doesn't actually seem to do anything, but I did connect it up.

PR #1161, opened 2018-07-17
PointDrawable: PointSprite is now correctly disabled in Core Profile.

emminizer:
Point Sprite was still not being deactivated because checkValidityOfAssociatedModes() wasn't being called on it.  It wasn't being called on it because GLObjectsVisitor does that, and the accept() that deals with the secondary stateset didn't account for GLObjectsVisitor.

Added a very targeted, small bit of code to account for this possibility.


emminizer:
Resolved instead by https://github.com/gwaldron/osgearth/commit/f3866680ac

PR #1162, opened 2018-07-19
Disable cancellation from mp

plevy:
Disable cancellation from mp to prevent flat elevation tiles.  Perhaps better to limit it to projected maps?


gwaldron:
Think Jason did this in b98eb16deb841483cfe78b3dd54a8f2ae4071b75

PR #1163, opened 2018-07-21
nullptr test added for the drawable parameter on entry to PrimitiveIntersector::intersect

damiandixon:
This was discovered when using old style picking. Occasionally drawable would be null.


gwaldron:
Appears to be an extraneous Qt-related change in here.

damiandixon:
Yes you are correct. I will have to do the pull request again. Sorry.

PR #1166, opened 2018-07-25
Earth Rewriter: Folder now based on absolute path instead of findData…

emminizer:
…File(), which can return empty string if file not found.  Supports stream-based loading and saving of Earth files instead of file-based.


emminizer:
We have a use case where a user wants to save off the current earth file with relative paths to a given location.  This normally works with files because the file exists and `findDataFile()` returns the input as given.  But when you're working with an `ostream`, there is not necessary a file there, but we still want to be able to get relative paths back from an existing directory.

This removes the `findDataFile()`, which was returning empty string in this case, instead relying on `getRealPath()` above to get the absolute path.  If that's good enough for the absolute filename, it should be good enough for the absolute filename's directory too.

PR #1169, opened 2018-07-27
Fixes to allow Qt related files to build after 'Annotations API refactor' changes.

damiandixon:
Updates to ensure Qt based code builds with latest core osgEarth changes.


gwaldron:
Thank you sir.

PR #1170, opened 2018-07-27
LGN: Render symbology now applies correctly on setStyle().  Fixed GL …

emminizer:
…Core Profile line width spam for extruded symbols.


emminizer:
Should be LGN (Local Geometry Node) not LGA (Local Geometry Annotation).  Oops.

gwaldron:
Yes lighting should definitely be on by default for extruded geometry

emminizer:
I amended the PR to enable lighting by default.  Thanks.

PR #1171, opened 2018-07-27
PointDrawable: glEnable/Disable(GL_POINT_SPRITE) is not supported for Core Profile

damiandixon:
GL_POINT_SPRITE is always enabled for Core Profile.

This modification stops OSG from calling glEnable/glDisable and
thus producing OpenGL errors when using Core Profile.

### DIFF
--- a/src/osgEarth/PointDrawable.cpp
+++ b/src/osgEarth/PointDrawable.cpp
@@ -362,7 +362,13 @@ PointDrawable::setupState()
         {
             _sharedStateSet = new osg::StateSet();
 
+#if defined(OSG_GL3_AVAILABLE) && !defined(OSG_GL2_AVAILABLE) && !defined(OSG_GL1_AVAILABLE)
+            // glEnable/Disable(GL_POINT_SPRITE) is not supported for Core Profile
+            // GL_POINT_SPRITE is always enabled for Core Profile
+            _sharedStateSet->setTextureAttributeAndModes(0, new osg::PointSprite(), osg::StateAttribute::OFF);
+#else
             _sharedStateSet->setTextureAttributeAndModes(0, new osg::PointSprite(), osg::StateAttribute::ON);
+#endif
 
             if (_gpu)
             {


emminizer:
Line 368 of your change sets `osg::PointSprite` to `OFF`.  This means the associated mode, `GL_POINT_SPRITE`, would still be set with `glDisable(GL_POINT_SPRITE)` somewhere during the frame.

The reason `osg::PointSprite` is here at all is to trigger the `state.setModeValidity(GL_POINT_SPRITE..., modeValid)` code in `osg::PointSprite::checkValidityOfAssociatedModes()`.  It also enables the `GL_POINT_SPRITE_COORD_ORIGIN` properly.  I guess in GL3 it doesn't matter too much if it's on or off, but I guess I don't see why it would be set to off?

Make sure you're running OSG 3.6.2.  https://github.com/openscenegraph/OpenSceneGraph/commit/d01edcb540a4 and https://github.com/openscenegraph/OpenSceneGraph/commit/500d3947e help with correct usage of Point Sprite in a GL3 Core Profile configuration and it can be messy to get point sprite working in both GL Core profile and Compatibility profile without it.

emminizer:
Sorry, in case it wasn't clear, `GL_POINT_SPRITE` mode can be set or unset as desired.  It is OSG's job to make sure that even if it's applied, it doesn't make it to the driver through a `glEnable()` or `glDisable()` job.  It is now doing its job correctly as of 3.6.2.

damiandixon:
I am building and running against OSG 3.6.2.
I'll have to take a look at OSG to see why GL_POINT_SPRITE glEnable/glDisable is occuring.
Thanks for pointing to where this is being checked.

damiandixon:
For Intel i965, Mesa 18.0.5, GL_ARB_point_sprite is supported.
However this does not mean that glEnable/glDisable is supported by the Core Profile for turning on/off point sprite. The standard says that point sprite is always enabled for Core Profile (hence the OFF in the patch).
This means that OSG is not performing sufficient checks. 
Probably better to update OSG, however I don't know my way around OSG particularly well below the api.

damiandixon:
I asked on the OSG mailing lists and Robert suggested the following:

> The glEnable/glDisable will be done directed by your scene graph via
> the StateSet::setMode(GLenum,..), so simply remove the
> setMode(GL_POINT_SPRITE).

Its going to probably be a couple of weeks before I get to look t this again.

emminizer:
I was able to reproduce your issue.  I created a small example and posted to OSG mailing list.  You can likely work around it in the meantime by adding an osg::PointSprite object to your Viewer's scene data.  I think that's the direction we'll be going in our tool to avoid the problem.

gwaldron:
Latest PointDrawable correctly calls checkValidityOfAssociatedModes in the PointSprite and works in GL CORE profile - going to close this out.

emminizer:
Should be fixed in OSG by https://github.com/openscenegraph/OpenSceneGraph/commit/e8b56c5c9 if you didn't already have a work-around, as per https://www.mail-archive.com/osg-users@lists.openscenegraph.org/msg77205.html

PR #1185, opened 2018-08-16
Triton: OSG built-in uniforms now correctly apply.  Correct program n…

emminizer:
…ow chosen per-camera for TritonDrawable uniforms.

### DIFF
--- a/src/osgEarth/NativeProgramAdapter
+++ b/src/osgEarth/NativeProgramAdapter
@@ -128,6 +128,16 @@ namespace osgEarth
                             useProgram = false;
                     }
                 }
+
+                // Apply each of the osg_*Matrix uniforms, which may be used by the native program
+                if (adapter.apply(_modelViewMatrixUniform.get(), useProgram))
+                    useProgram = false;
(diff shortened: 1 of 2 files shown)


emminizer:
There was a mistake in https://github.com/gwaldron/osgearth/commit/83c1ce1f88 that breaks Triton >= 4, but fixed Triton < 4.  Because `Version.h` was not included (from Triton), the define was never set, so new Triton (4+) is pulling out variables from the wrong program.  This means Triton 4+ never carries over Uniforms; this is now fixed by including `Version.h`.

The other problem is that Triton shaders might depend on uniforms like `osg_ProjectionMatrix` in order to detect things like ortho display in shader.  While `NativeProgramAdapter` correctly found the uniform ID mapping for `osg_ProjectionMatrix`, it never actually made it to Triton because `State::applyModelViewAndProjectionUinformsIfRequired()` is the code that actually applies those uniforms to the `State`, and it isn't run here.  Calling it directly is not appropriate because the `_lastAppliedProgramObject` is not set.  This change just hits all of the built-ins explicitly.  If the Triton user functions don't use it, there's no change in behavior.  If they do use one, it gets applied.

PR #1195, opened 2018-08-28
Topic gles3

tomhog:
Improved android and iOS support for gles3


src/osgEarth/ShaderGenerator.cpp
```diff
@@ -701,7 +702,9 @@ ShaderGenerator::apply(osg::ClipNode& node)
         "void oe_sg_clipnode(inout vec4 vertex_view)\n"
         "{\n"
         "    vec4 plane_view = gl_ModelViewMatrix * oe_sg_clipnode_plane_model; \n"
+        "#ifndef GL_ES'n"
```
gwaldron: Typo?

src/osgEarth/ObjectIndex.cpp
```diff
@@ -63,7 +63,11 @@ ObjectIndex::ObjectIndex() :
 _idGen( STARTING_OBJECT_ID )
 {
     _attribName     = "oe_index_objectid_attr";
+#if defined(OSG_GLES3_AVAILABLE)
+    _attribLocation = osg::Drawable::TEXTURE_COORDS_5; // Why is SECONDARY_COLORS occupied. Could it actually be a QT issue?
```
gwaldron: This is an issue on OSX as well -- also happening on Android/IOS?

gwaldron: This is also an unresolved issue on OSX. Does it also happen on IOS and Android?

src/osgEarth/LineDrawable.cpp
```diff
@@ -1159,9 +1159,17 @@ LineDrawable::setupShaders()
             vp->addBindAttribLocation("oe_LineDrawable_prev", LineDrawable::PreviousVertexAttrLocation);
             vp->addBindAttribLocation("oe_LineDrawable_next", LineDrawable::NextVertexAttrLocation);
             ss->getOrCreateUniform("oe_LineDrawable_limits", osg::Uniform::FLOAT_VEC2)->set(osg::Vec2f(-1,-1));
-            //_gpuStateSet->getOrCreateUniform("oe_GL_LineWidth", osg::Uniform::FLOAT)->set(1.0f);
+            //_gpuStateSet->getOrCreateUniform("oe_GL_LineWidth", osg::Uniform::FLOAT)->set(30.0f);
             //_gpuStateSet->getOrCreateUniform("oe_GL_LineStippleFactor", osg::Uniform::INT)->set(1);
             //_gpuStateSet->getOrCreateUniform("oe_GL_LineStipplePattern", osg::Uniform::INT)->set((int)~0);
+            
+#if defined(OSG_GLES3_AVAILABLE)
```
gwaldron: The expectation here is that the coder will call GLUtils::setGlobalDefaults() someone else in the application, which initializes this uniform's default value.

src/osgEarth/Capabilities.cpp
```diff
@@ -159,10 +159,19 @@ _isCoreProfile          ( true )
     // create a graphics context so we can query OpenGL support:
     MyGraphicsContext mgc;
 
+#ifdef __ANDROID__
```
gwaldron: Some comments explaining these conditionals would be helpful.

src/applications/osgearth_viewerIOS/osgPlugins.h
```diff
@@ -180,7 +156,7 @@ USE_OSGEARTH_SIMPLE_SYMBOL(coverage)
 USE_OSGEARTH_SIMPLE_SYMBOL(extrusion)
 USE_OSGEARTH_SIMPLE_SYMBOL(icon)
 USE_OSGEARTH_SIMPLE_SYMBOL(line)
-USE_OSGEARTH_SIMPLE_SYMBOL(marker)
+//USE_OSGEARTH_SIMPLE_SYMBOL(marker) // why does this one not link, seems to have a register function
```
gwaldron: It's ok, marker no longer exists

src/osgEarth/VirtualProgram.cpp
```diff
@@ -291,13 +291,19 @@ namespace
     }
 
 
-    void parseShaderForMerging( const std::string& source, unsigned& version, std::string& subversion, HeaderMap& precisions, HeaderMap& headers, std::stringstream& body )
```
gwaldron: Down the road we should break the shader merging code out into ShaderUtils or something.

src/osgEarth/ObjectIndex.cpp
remoe: Glenn, this is solved? Here: https://github.com/gwaldron/osgearth/issues/1098

gwaldron: Oh right ... forgot about that. Thanks Remo :)

src/osgEarth/Capabilities.cpp
tomhog: Will do, basically we can’t currently create contexts on the native side with android so this skips that part.

src/osgEarth/LineDrawable.cpp
tomhog: I think this may have come from me working on this at the same time as you were working on the line code. Now it’s settled down i will review.

src/osgEarth/ObjectIndex.cpp
tomhog: Good to hear i was going to ask as it seemed this  might be adustable from the application level also

src/osgEarth/ShaderGenerator.cpp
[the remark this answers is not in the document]
tomhog: Must be, im supprised that still worked

src/osgEarth/VirtualProgram.cpp
tomhog: 100% now osg has the condintional compilation of shaders the merging has become way more complicated. This will trip over in certain cases. I’ve informed chris that a rewrite is a better approach then hacking a fix everytime a shadet combination fails to compile due to bad merging.

tomhog:
Thanks for looking at these so swiftly Glenn, been quite a lot of work getting both android and iOS playing nicely and had got to the point we just needed to get it merged before it fell to the way side.

tomhog:
Any chance of getting this in for 2.10 @gwaldron ?

gwaldron:
Tom,
I added a ShaderMerger class, but I did not activate it for GLES because I don't have a test platform. Do you have time to give it a try and see if there are any issues? The diverging code starts here:
https://github.com/gwaldron/osgearth/blob/master/src/osgEarth/VirtualProgram.cpp#L607

PR #1197, opened 2018-08-30
Use the flag image from data folder

plevy:


gwaldron:
Fixed by replacing the texture.

PR #1205, opened 2018-09-20
Capabilities: Better core profile test based on results from ATI card…

emminizer:
… testing.


emminizer:
The original test worked great on Intel, Mesa, and NVidia from my testing.  However, we found an older ATI card that still had that extension marked as available even in Core Profile.  This change works on all 4 driver architectures that I was able to test.

PR #1211, opened 2018-09-26
Add level for model placement

plevy:
Add level for model placement


gwaldron:
fixed it a different way - thanks

PR #1212, opened 2018-09-29
-fix incorrect iteration of component iterator

pprabhu78:


gwaldron:
Good find, thanks Poojan

pprabhu78:
Thanks Glenn!

PR #1216, opened 2018-10-08
Fix compilation error with Geos 3.7.0.

Hallot:
According to the 3.7.0 release notes, the C++ API changed, resulting in a compilation error:
osgearth/src/osgEarthSymbology/GEOS:51:38: error: ‘unique_ptr’ in ‘class geos::geom::GeometryFactory’ does not name a type.

Instead, use geos::geom::GeometryFactory::Ptr which should be compatible with Geos 3.6 as well.

### DIFF
--- a/src/osgEarthSymbology/GEOS
+++ b/src/osgEarthSymbology/GEOS
@@ -47,7 +47,9 @@ namespace osgEarth { namespace Symbology
         void disposeGeometry(geos::geom::Geometry* input);
 
     protected:
-#if GEOS_VERSION_MAJOR >= 3 && GEOS_VERSION_MINOR >= 6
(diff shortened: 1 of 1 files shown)


gwaldron:
Failed to build with AppVeyor CI -- please check that you got the version checks right.

Hallot:
I was mistaken that it would also work for Geos 3.6.x.
I kept the check for 3.6 only, and added the new one for >= 3.7.

Hallot:
@gwaldron I am not sure why Travis failed, but it does not seem to be related to my code.
Any way to restart it? (It's the same for the other PR)

gwaldron:
You're right, Travis fail is unrelated. Thanks.

PR #1217, opened 2018-10-08
Fix signed integer overflow when computing the threshold.

Hallot:
osgearth/src/osgEarth/JsonUtils.cpp:2483:54: warning: integer overflow in expression of type ‘osgEarth::Json::Value::Int’ {aka ‘int’} results in ‘-2147483648’ [-Woverflow]
    Value::UInt threshold = (isNegative ? Value::UInt(-Value::minInt)

For signed integers, -minInt = maxInt + 1.
This results in an overflow, which is undefined behaviour for signed integers.

Moreover, on most machines, the overflow would wrap around, resulting in the threshold being minInt for negative numbers.
Negative values would always be decoded as double.

### DIFF
--- a/src/osgEarth/JsonUtils.cpp
+++ b/src/osgEarth/JsonUtils.cpp
@@ -2480,7 +2480,7 @@ Reader::decodeNumber( Token &token )
    bool isNegative = *current == '-';
    if ( isNegative )
       ++current;
-   Value::UInt threshold = (isNegative ? Value::UInt(-Value::minInt) 
+   Value::UInt threshold = (isNegative ? Value::UInt(Value::maxInt)
                                        : Value::maxUInt) / 10;
    Value::UInt value = 0;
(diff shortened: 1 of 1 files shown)


gwaldron:
Are you actually seeing a bug? Or just trying to squash a compiler warning? We didn't write this code, so we are hesitant to mess with it unless there is a real problem.

Hallot:
> Are you actually seeing a bug? Or just trying to squash a compiler warning? We didn't write this code, so we are hesitant to mess with it unless there is a real problem.

Just the compiler warning.  
I don't think it's easily observable, since the only result of this bug is to decode the value as double, which (I haven't checked) should work just fine.  
It's potentially slower, and there might be some precision issues, but ultimately minor.

I understand if you do not want to modify the code without a concrete bug as this introduces risk.
It might be worth it if it is a hot code path though. Your call.

PR #1224, opened 2018-10-12
-changes for API changes with regards latest SilverLining

pprabhu78:
Hi Glenn,

Here's a patch against the latest SilverLining that take into account the API changes mentioned here:

https://sundog-soft.com/2018/10/important-api-updates-in-silverlining-5-30/

Its minor, but important

### DIFF
--- a/src/osgEarthSilverLining/SilverLiningAPIWrapper
+++ b/src/osgEarthSilverLining/SilverLiningAPIWrapper
@@ -199,7 +199,7 @@ namespace osgEarth { namespace SilverLining
         bool GetFogEnabled() const; 
         void GetFogSettings(float &density, osg::Vec3 &color) const;
         void GetHorizonColor(float yaw, float pitch, osg::Vec3 &color) const;
-        bool GetEnvironmentMap(GLuint &environmentMapTex, int facesToRender = 6, bool floatingPoint = false, void* cameraID = 0,bool drawClouds = true, bool drawSunAndMoon = true) const;
+        bool GetEnvironmentMap(GLuint &environmentMapTex, int facesToRender = 6, bool floatingPoint = false, void* cameraID = 0, bool drawClouds = true, bool drawSunAndMoon = true, bool geocentricMode = false) const;
         void SetHaze(const osg::Vec3 &color, double depth, double density);
     public:
         Atmosphere(uintptr_t handle) : _handle(handle) { }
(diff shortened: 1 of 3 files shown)


gwaldron:
Thx - do you happen to know if these changes backwards compatible?

pprabhu78:
They are not backward compatible.  With older versions,  you would have to
remove the last parameter passed in.

emminizer:
At least in SilverLining 5.009 I agree that it's not compatible; and unfortunately there doesn't seem to be an easy way to do a version check with an ifdef, that I can see.  That's unfortunate.

pprabhu78:
The api changes add default parameters. They would be compatible if it was
not geocentric mode, which it is in the for osgearth.

Not sure whether SL version is maintained somehow in SL code. I copied
Frank on this email.

On Oct 15, 2018 5:52 AM, "Daniel Emminizer" <notifications@github.com>
wrote:

At least in SilverLining 5.009 I agree that it's not compatible; and
unfortunately there doesn't seem to be an easy way to do a version check
with an ifdef, that I can see. That's unfortunate.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/gwaldron/osgearth/pull/1224#issuecomment-429780251>,
or mute
the thread

pprabhu78:
There is a Version.h header in SilverLining's public headers, where
SILVERLINING_MAJOR_VERSION and SILVERLINING_MINOR_VERSION are defined.

Thank you,

Frank Kane
Founder
Sundog Software, LLC

http://www.sundog-soft.com/

http://www.linkedin.com/pub/frank-kane/17/434/764

gwaldron:
Dan, can you double check that? I seem to have version 5.005 (or 006), and I do have a Version.h file.

emminizer:
Double checked.  It's not in my Program Files's SilverLining SDK (Full source) directory.  I checked Public Headers there and it's not there.  I do see a Version.h in my Linux build.  I did an uninstall and a reinstall.  I confirmed I'm on 5.009.  The Full source install does not include a Version.h.  I checked Public Headers directory, which has 25 headers, none including Version.h.  A full grep -r through the extracted exe in Program Files does not include the text `SILVERLINING_MAJOR_VERSION` anywhere.

I'll pass along a note to their team to indicate the problem.  From my side I'm fine with this change here, if you don't mind trying to conditionally compile the calls to `DrawObjects()` and `GetEnvironmentMap()` with preprocessor conditionals?  We're happy to update our SilverLining version, it's just a matter of getting the time to do so since we have to rebuild without DX support.

I'm guessing this is a bug with the Windows installer.  I can limp along by copying the one from Linux over since they're the same version.

pprabhu78:
I copied Frank on this reply.

pprabhu78:
It's certainly possible that there was some sort of build or installer
glitch at that time that we never knew about. If so I apologize for the
oversight.

However, I can confirm that Version.h is being generated reliably now. The
solution of just copying over the Version.h from the Linux distribution
seems fine to me. But, a conditional compile of the API changes that
depends not only on the value of our version #define's but also their
existence would do the job too. If they don't exist at all, it's safe to
say you're using a SilverLining version from before the API changes.

Thank you,

Frank Kane
Founder
Sundog Software, LLC

http://www.sundog-soft.com/

http://www.linkedin.com/pub/frank-kane/17/434/764

gwaldron:
Implemented with version checking in 6a647eeb50457af12076b0ac27f5d012bc1d4249

pprabhu78:
Copying Frank

pprabhu78:
Copying Frank

PR #1243, opened 2018-11-15
Replace the osgEarth/Version file with file generated cmake

Vic-Min:
To prevent mismatch of version numbers specified in CMakeLists.txt and osgEarth/Version


gwaldron:
Done in the version 3.5

PR #1252, opened 2018-11-28
Fix bounds checking in RTTPicker::pick

### LINKED ISSUE #1250: Bounds check in RTTPicker::pick is incomplete
Currently, in `RTTPicker::pick` I see the [following code](https://github.com/gwaldron/osgearth/blob/master/src/osgEarthUtil/RTTPicker.cpp#L306):
```
    // check the bounds:
    if ( u < 0.0f || u > 1.0f || v < 0.0f || v > 1.0f )
        return false;
```
This misses the cases when `u==1 || v==1`. When this happens, in particular, for `v`, `SpiralIterator` will read past the end of the image data, because `_offsetY` will be equal to `h`, while the largest possible offset is `h-1` since it's indexed from 0.

This in particular happens with OpenSceneGraph before its [commit abd2fa53c4acede9acbca536c11922cfd9fcb00f](https://github.com/openscenegraph/OpenSceneGraph/commit/abd2fa53c4acede9acbca536c11922cfd9fcb00f), which sometimes lets `GUIEventAdapter::getY()` return height (when mouse cursor is at the topmost pixel).

The result is that we sometimes (when the data read appears >0 in `RTTPicker::checkForPickResult`) get garbage `ObjectID` in the `Callback::onHit()` implementation, which could also potentially collide with an existing id, resulting in false pick.

Same could happen for `u` coordinate. I don't how to reproduce this case, since there hasn't been any off-by-one bug in OSG (due to X coordinate not being inverted), but this is possible in `RTTPicker` anyway and should be fixed.

10110111:
This makes sure we don't read at _offsetX==w nor at _offsetY==h in
SpiralIterator::next(), which would give garbage results.
This fixes #1250.

### DIFF
--- a/src/osgEarthUtil/RTTPicker.cpp
+++ b/src/osgEarthUtil/RTTPicker.cpp
@@ -303,7 +303,7 @@ RTTPicker::pick(osg::View* view, float mouseX, float mouseY, Callback* callback)
     float v = (mouseY - (float)vp->y())/(float)vp->height();
 
     // check the bounds:
-    if ( u < 0.0f || u > 1.0f || v < 0.0f || v > 1.0f )
+    if ( u < 0.0f || u >= 1.0f || v < 0.0f || v >= 1.0f )
         return false;
 
     // install the RTT pick camera under this view's camera if it's not already:


gwaldron:
I will commit this to the 2.10 branch as well. Thanks.

PR #1253, opened 2018-11-29
ImageLayer: Callbacks now fire when altitude changes.

emminizer:


gwaldron:
Plan was ultimately to remove all these callbacks, but for now I will merge this for convenience sake :)

PR #1266, opened 2019-01-07
fix tree transparency

plevy:


gwaldron:
Example is no longer in the repo

PR #1299, opened 2019-03-07
Change http to https connections for Bing driver.

Hallot:


gwaldron:
Also fixed in the 3.0 branch

PR #1312, opened 2019-04-18
Fix Config:: _refMap lost when doing Config::merge

andyzhshg:
`Config:: _refMap` is lost when doing `Config::merge`. I find this issue when running `osgearth_features` demo with `--mem` option.


gwaldron:
Thanks!

PR #1315, opened 2019-04-30
webp plugin

remoe:
WebP plugin. Code is based on WebP 0.6.1.


gwaldron:
Thanks Remo!
It doesn't look like you are testing the WEBP_FOUND variable in the CMake setup. Can you add that?

remoe:
done.

PR #1327, opened 2019-07-05
Fix build errors on GCC 6.3

Himbeertoni:


Himbeertoni:
I don't know why the automatic build failed, the modifications I did have nothing to do with the build error.

PR #1332, opened 2019-07-12
fix: optional webp plugin

remoe:
WebP plugin guard was wrong.


remoe:
Glenn, sorry I was to late for 2.10.2 :/

remoe:
I've tested it on a full osgEarth Linux-Ubuntu build with WebP 1.0.2.

gwaldron:
I snuck it in :)

PR #1333, opened 2019-07-17
2.8 cmake fixes

rhabacker:
These cmake-related patches were required by various applications that use osgearth.


rhabacker:
> continuous-integration/appveyor/pr — AppVeyor build failed 

This issue is unrelated to this patch.

gwaldron:
Closing. Most of these are obsolete now with the possible exception of the "WIN32 and not MSVC" directive but will need more information before using that

PR #1335, opened 2019-07-19
Activate viewpoints ui if the earth file was not loaded by the osgearth_viewer

640kb:
If the Earth file is loaded with another viewer (e.g. osgviewer), the UI must be added so that the viewpoints can be used.

### DIFF
--- a/src/osgEarthDrivers/viewpoints/ViewpointsExtension
+++ b/src/osgEarthDrivers/viewpoints/ViewpointsExtension
@@ -21,6 +21,7 @@
 
 #include <osgEarth/Extension>
 #include <osgEarth/Viewpoint>
+#include <osgEarth/MapNode>
 #include <osgEarthUtil/Controls>
 #include <osgGA/GUIEventAdapter>
 #include <osg/View>
@@ -38,6 +39,7 @@ namespace osgEarth { namespace Viewpoints
     class ViewpointsExtension : public Extension,
                                 public ExtensionInterface<osg::View>,
                                 public ExtensionInterface<Control>,
+                                public ExtensionInterface<MapNode>,
                                 public ConfigOptions
     {
     public:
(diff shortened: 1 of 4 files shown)


remoe:
Hi, sorry but a driver "earth/ReaderWriterOsgEarth.cpp" has nothing to do with UI. It's wrong when one include this:

```
#include <osgEarthUtil/Controls>
#include <osgEarthUtil/ExampleResources>
#include <osgEarthSymbology/Color>
#include <osgViewer/Viewer>
```

640kb:
I see what you mean, I already have an idea how I could work this over.

640kb:
Okay, I think I fixed your mentioned problem.

- I moved the code to the driver viewpoints
- Viewpoints implements now an additional ExtensionInterface with the class MapNode as parameter
- The template interface gets called by the earth driver, where the event handler can be added to the map node.
- The event handler adds the ui for the viewpoints if it does not already exist.

gwaldron:
The reason that osgearth_viewer (et al) loads the Viewpoints UI is because of [this code block](https://github.com/gwaldron/osgearth/blob/master/src/osgEarthUtil/ExampleResources.cpp#L645) in ExampleResources.cpp. If you do this in your own app you will get Extension user interfaces. If that what you're looking for?

640kb:
I've seen the code, but I think that's only a solution if you have the ability to customize the viewer and also have access to the code. And the adjustments would have to be done once with every new or different viewer.

So wouldn't it make sense if the plugin integrated it?

gwaldron:
Maybe I am having trouble understanding your use case. Are you trying to add a viewpoints UI component to a 3rd party application?

640kb:
That's right. If I load an earth file with a 3rd party application, the ui is missing and the viewpoints are also not usable.

640kb:
I fixed a little issue. It is now checked whether an EarthManipulator is already available before a new camera manipulator is set.

gwaldron:
Closing. The Viewpoints plugin is meant to support the examples; if you wish to use it with a custom app you can use the Extension interface as described

PR #1343, opened 2019-08-14
Triton: Turn off UBO if Mesa driver detected to prevent GL errors.

emminizer:
While testing on Linux, I'm getting a constant stream of errors on console when using Triton.  I traced it down to Unified Buffer Objects (UBO) using apitrace and MESA_DEBUG flag.  The UBO is on by default in latest Triton.config file (not sure about previous values).  Once I turned it off, the osgEarth and SIMDIS SDK Triton demos both worked well.

Note -- It might be advantageous to expose an interface to set individual Configuration settings both for Triton (and SilverLining for that matter) by creating a new field in TritonOptions.  I did not go that far with this approach.  I also did not attempt to fix this in Triton code directly because presumably eventually Mesa drivers might support UBO, and Triton already has a method for configuring this (Triton.config, Environment.SetConfigOption).

Testing was only done on Linux side with latest master of osgEarth and latest Triton.  Testing was done in osgearth_triton, example_ocean in SIMDIS SDK, and in SIMDIS itself.

### DIFF
--- a/src/osgEarthTriton/TritonContext.cpp
+++ b/src/osgEarthTriton/TritonContext.cpp
@@ -20,6 +20,8 @@
 #include <osg/GLExtensions>
 #include <osg/Math>
 #include <osgDB/FileNameUtils>
+#include <osgEarth/Capabilities>
+#include <osgEarth/Registry>
 #include <osgEarth/SpatialReference>
 #include <cstdlib>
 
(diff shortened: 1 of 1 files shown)


emminizer:
Disregard this PR; during testing I apparently still had my Triton.config file set to "no" for this option.  Upon rebuild of everything I noticed this did not have desired effect.

Turns out this is read by the configuration from the Triton.config in Environment's initialize(), overriding any value you set here, then immediately used before control is returned to the user.  The only solutions that would work are:
* Edit the Triton.config manually ahead of time
* Override a method in ResourceLoader to force the value immediately after resource file loaded (messy)

There's not a good solution on osgEarth side of the fence.

PR #1346, opened 2019-08-15
ogr: support TRIANGLEZ and TINZ

remoe:
TRAINGLEZ and TINZ are only synonyms:

TRIANGLEZ -> Polygon
TINZ -> MultiPolygon


remoe:
jason, had integrated TIN support.

PR #1351, opened 2019-08-23
Color: Using statement instead of typedef, to fix MSVC 2010-2015 errors

emminizer:


emminizer:
Fixes issue introduced in https://github.com/gwaldron/osgearth/commit/fcf48d5c1b that impacts MSVC 2010, 2013, 2015.  I tested this fix on MSVC 2010 and MSVC 2019 (to make sure it works on our oldest and newest supported compilers), and also tested on g++ 4.4.  All 3 compilers appeared to accept this.

PR #1356, opened 2019-09-11
Picking with cascaded draping needs 4 frames?

remoe:
Picking with cascaded draping doesn't work on first hit, because the picker hasn't enough frames?
I can't reproduce it after this fix. But is it correct?


gwaldron:
I was not able to reproduce this issue with osgearth_pick. Are you? Maybe you are rendering to an FBO and this is causing the issue?

remoe:
Ok, I need to test more.

PR #1360, opened 2019-09-19
static QString use tr()  will failed.

qiup-Q:
BTW, it will crash , when i run applications/osgearth_qt_windows on Windows 10 with Qt 5.12.3 and debug version.
https://stackoverflow.com/questions/1486492/qt-tr-does-not-seem-to-work-on-static-constant-members


gwaldron:
osgEarthQt is no longer supported.

PR #1367, opened 2019-10-01
Fix build with gcc-6 at Debian-9

drons:


gwaldron:
Fixes manually integrated

PR #1374, opened 2019-10-27
Some clang-tidy fixies

asapelkin:
1) Used single quotes for single character in string::find (more efficient)
2) Removed unnecessary copy initialization in some places


gwaldron:
Manually applied most of these to main branch.

PR #1391, opened 2019-11-18
add useOSGTessellator to select the tessellator

remoe:
add "useOSGTessellator" to select the tessellator for the geometry compiler in the feature layer


jasonbeverage:
Hi Remo,

I'm merging this into the 2.10 branch, and I'm also going to go ahead and merge it with the tin_support branch as well so we can get support for your 3d geometries.

PR #1415, opened 2019-12-18
bugfix: incorrect WMS URI when using wrong locale

boitumeloruf:
I have integrated the changes of podsvirov as discussed in https://github.com/gwaldron/osgearth/pull/835#issuecomment-567031484.

I have tested it on a machine with a german locale. Before the fix, decimal numbers of the BBX were written with a ','. After fix the URI is created with a '.' in the decimal numbers.

### DIFF
--- a/src/osgEarthDrivers/wms/ReaderWriterWMS.cpp
+++ b/src/osgEarthDrivers/wms/ReaderWriterWMS.cpp
@@ -430,9 +430,16 @@ public:
     {
         double minx, miny, maxx, maxy;
         key.getExtent().getBounds( minx, miny, maxx, maxy);
+	
+	// Convert numbers only in "C" locale (decimal point is dot)
+        // Store current numeric locale
+        char *cur_lc_numeric = setlocale(LC_NUMERIC, "C");
         
         char buf[2048];
         sprintf(buf, _prototype.c_str(), minx, miny, maxx, maxy);
+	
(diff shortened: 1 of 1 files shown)


tompollok:
I also have issues with MWS on my side. This patch seems to fix it. Whats blocking this PR from merge?

gwaldron:
setlocale is not thread-safe; in fact it is a per-process function that will affect all running threads. Is there some reason you cannot just call setlocale at application start?

boitumeloruf:
I understand that it is not thread-safe. However, we are developing applications in german and want to use the appropriate locale for display (e.g. have ',' as decimal separator) but when accessing WMS it requires '.' to be the decimal separator. Hence, it is not possible to set the locale for the whole application. Is there another option to fix this bug?

gwaldron:
Does this apply just to WMS or to other protocols as well?
Would imbue be a solution? https://en.cppreference.com/w/cpp/io/basic_ios/imbue

boitumeloruf:
Thank you. I have only used WMS so far, so i don't know about other protocols.

Using imbue seems to be an alternative. I am currently still on version 2.10. However, I will test it with imbue and let you know.

gwaldron:
Using ios::imbue in master/3.1. If you wish to submit an imbue patch to the 2.10 branch, please do so in a new PR. Thanks!

PR #1427, opened 2020-01-17
KML_Geometry parseCoords ensures all coords are parsed in case of a comment within the coord list

geissel1:
If a comment is found while parsing the coordinate list in the rapidxml.hpp, the coords found after the comment are all added as another child node, but the m_value field string only reflects the first node. Instead, iterate through all of the coordinate node's child nodes to make sure all coordinates are parsed.


gwaldron:
Patrick, did you mean to PR this on the 3.0 branch?

geissel1:
I may be out of date, is there a different repo that is more current?

PR #1442, opened 2020-02-06
Tracy CMake update?

remoe:
I've build Tracy from source (MacOSX) and tried to build a sample app. I haven't found an install make. @jasonbeverage, do you have found one? So, I've adapted it for the following installation:

```
lib/libtracy-release.so
include/TracyClient/Tracy.hpp
```

### DIFF
--- a/CMakeModules/FindTracy.cmake
+++ b/CMakeModules/FindTracy.cmake
@@ -6,7 +6,7 @@
 
 SET(TRACY_DIR "" CACHE PATH "Root directory of Tracy distribution")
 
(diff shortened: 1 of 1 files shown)


gwaldron:
Thank Remo, that finder was an oversight when we changed the include to TracyClient/Tracy.hpp.

jasonbeverage:
Hey Remo, the TracyClient is a name of a client DLL project I came up with, so there is nothing official there.  For now this will work, but it might change in the future FYI.  Nothing too drastic though :)

Issue #1440, opened 2020-02-05
MaskGenerator: figure out "duplicate point" and "trajd" errors.


gwaldron:
The MG dumps these out occasionally (just run mask.earth to see the dp errors) and it is likely a precision loss error. I suspect the osg delaunay triangulator, which only runs in single precision, might be the cause. Run mask.earth to see the "clipped corner" issue as well, which might (hopefully) be related. Do we need a double-precision version of the DT? Or is the problem elsewhere?

gwaldron:
Looks good :)

PR #1448, opened 2020-02-11
Moore/mask

timoore:
Fix duplicate vertices and clipped corners in tile masks.

https://github.com/gwaldron/osgearth/issues/1440


timoore:
Screwed up; trying again

PR #1459, opened 2020-02-28
Moved PrimiveIntersector::Intersection copy constructor

Moohasha:
The copy constructor for PrimitiveIntersector::Intersection was defined in
the cpp file while the rest of the struct was defined inline in the
header.  This meant that you couldn't copy the struct without from your
own project without getting linker errors.  I simply moved the copy ctor
to be inline like the rest of the struct.


gwaldron:
Are you on Linux? Adding OSGEARTH_EXPORT would prob work too. Thanks

PR #1464, opened 2020-03-03
LayerDrawable: Avoid possible buffer overflow from sprintf() usage.

emminizer:
If layer name is unusually long, this could overflow the `buf[64]`.  `Stringify` prevents that.


emminizer:
Closing, fixed instead by https://github.com/gwaldron/osgearth/commit/c43ffa37845

PR #1479, opened 2020-04-04
fix std::max、std::min error C2589 on windows

lcp0578:
fix std::max、std::min error C2589 on windows


remoe:
I've built osgEarth 3.0 with VC2017, VC2019 and XCode 11. Don't have this issue.

lcp0578:
@remoe Oh，I will check my env and close it.

Issue #1491, opened 2020-04-16
Sky: does not work from earth file


gwaldron:
<sky> does not connect to the View; space looks good but none of the atmospheric lighting works. Command line --sky still works properly.

timoore:
Fixed by pull #1495

PR #1495, opened 2020-04-18
fix: Sky does not work from earth file

remoe:
solves #1491


timoore:
I'm happy to check this in, but it look like there was an effort in 3.0 to take the sky option out of .earth files; only one .earth file in tests/ has the option, and the instructions were changed to say "run with --sky option." Glenn?

Issue #1455, opened 2020-02-20
(GEOS 3.8) Geos function mismatch when building osgEarthSymbology/GEOS.cpp


paul-jewell:
I am receiving the error shown below when building with release 2.10.1. Geos version installed is 3.8.0-r1. I have replicated this on two different computer systems, both running Gentoo. Have you seen any errors similar to this? 

I appreciate your comments!

Error message:
`[ 28%] Building CXX object src/osgEarthSymbology/CMakeFiles/osgEarthSymbology.dir/GeometryRasterizer.cpp.o
cd /var/tmp/portage/sci-geosciences/osgearth-2.10.1/work/osgearth-2.10.1_build/src/osgEarthSymbology && /usr/bin/x86_64-pc-linux-gnu-g++ -DOSGEARTHSYMBOLOGY_LIBRARY -DOSGEARTH_CXX11 -DOSGEARTH_HAVE_GEOS -DosgEarthSymbology_EXPORTS -I/var/tmp/portage/sci-geosciences/osgearth-2.10.1/work/osgearth-osgearth-2.10.1/src -I/var/tmp/portage/sci-geosciences/osgearth-2.10.1/work/osgearth-osgearth-2.10.1   -DNDEBUG -march=native -O2 -pipe -fPIC   -std=gnu++11 -o CMakeFiles/osgEarthSymbology.dir/GeometryRasterizer.cpp.o -c /var/tmp/portage/sci-geosciences/osgearth-2.10.1/work/osgearth-osgearth-2.10.1/src/osgEarthSymbology/GeometryRasterizer.cpp
/var/tmp/portage/sci-geosciences/osgearth-2.10.1/work/osgearth-osgearth-2.10.1/src/osgEarthSymbology/GEOS.cpp: In function ‘geos::geom::CoordinateSequence* {anonymous}::vec3dArray2CoordSeq(const osgEarth::Symbology::Geometry*, bool, const geos::geom::CoordinateSequenceFactory*)’:
/var/tmp/portage/sci-geosciences/osgearth-2.10.1/work/osgearth-osgearth-2.10.1/src/osgEarthSymbology/GEOS.cpp:67:56: error: cannot convert ‘std::unique_ptr<geos::geom::CoordinateSequence>’ to ‘geos::geom::CoordinateSequence*’ in initialization
   67 |         geom::CoordinateSequence* seq = factory->create( coords );
      |                                         ~~~~~~~~~~~~~~~^~~~~~~~~~
[...]

Issue #1378, opened 2019-10-29
Can't compile with GEOS 3.8.0


lbartoletti:
GEOS 3.8.0 was release 3 weeks ago.

osgEarth can't compile with it. Here an error log:
http://package18.nyi.freebsd.org/data/120amd64-default-PR238988/2019-10-22_05h46m02s/logs/errors/osgearth-2.10.1_3.log

Thank you.

sgarfinkel:
@gwaldron Is there a workaround to this? Both VCPKG and Homebrew currently link to GEOS 3.8.0 which makes building OSGEarth irritating.

bosth:
Been a while since I did any work in C++, but I knocked this patch together: https://git.slackbuilds.org/slackbuilds/tree/gis/osgEarth/geos-3_8-support.patch

lbartoletti:
@bosth Thanks. LGTM

gwaldron:
GEOS 3.8.x works

PR #1497, opened 2020-04-19
GEOS 3.8 compile fix

remoe:
fix for https://github.com/gwaldron/osgearth/issues/1455 and https://github.com/gwaldron/osgearth/issues/1378
for osgEarth 3.0

### DIFF
--- a/src/osgEarth/GEOS.cpp
+++ b/src/osgEarth/GEOS.cpp
@@ -63,9 +63,13 @@ namespace
         {
             coords->push_back( coords->front() );
         }
+#if GEOS_VERSION_AT_LEAST(3,8)
+        geom::CoordinateSequence::Ptr seq = factory->create(coords);
+        return seq.get();
+#else
         geom::CoordinateSequence* seq = factory->create( coords );
-
         return seq;
+#endif
     }
 
     geom::Geometry*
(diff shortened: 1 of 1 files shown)


src/osgEarth/GEOS.cpp
```diff
@@ -63,9 +63,13 @@ namespace
         {
             coords->push_back( coords->front() );
         }
+#if GEOS_VERSION_AT_LEAST(3,8)
+        geom::CoordinateSequence::Ptr seq = factory->create(coords);
+        return seq.get();
```
ankurverma85, line 68: This seems wrong. Ptr is a unique_ptr so the memory will be freed up as soon as the function is returned.
This function  would be returning freed up memory.

seq.release is probably what would work here but there migh be a memory leak

ankurverma85:
osgearth seems broken after the geos 3.8.1 update.
I'm curious as to how did the geos update sneak in (which breaks osgearth) without triggering a pipeline build failure in osgearth.
Shouldnt this have been caught by the pipeline build in the Pull-Request for the geos 3.8.1 update ?

jasonbeverage:
I just pushed a change to build against the geos c API which should be fine
against 3.8.   Make sure you link against the geos_c.lib and not the
geos.lib when you rebuild.

PR #1507, opened 2020-04-28
WindLayer: Fix GL3 build on GLintptr compile error.

emminizer:
Attempting to fix issue pointed out by https://github.com/gwaldron/osgearth/commit/896debae11dfc4c9d26e63ae5d3bbbcf82e8b3ae#r38795634

I do not have a system where the osg::GLintptr is defined on which to test this however, I only have the GL3 system where that does not compile and this changeset does.  Tested on Linux gcc 8.3.1 with what looks like `-std=gnu++11`

### DIFF
--- a/src/osgEarth/WindLayer.cpp
+++ b/src/osgEarth/WindLayer.cpp
@@ -48,6 +48,11 @@ namespace
         GLfloat speed;
     };
 
+    // osg::GLintptr is not defined on all systems, and GLintptr is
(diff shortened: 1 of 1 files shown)


gwaldron:
Dan, thanks, but I am just going to replace it with a GLint :)

emminizer:
No problem.  I did not inspect usage.  Primarily concerned on making build work.  Based on usage I agree `GLuint` is better.  Thanks for fixing.

PR #1510, opened 2020-04-30
Fix osg::Observer usage.  Removed Observer from AnnotationNode, fixed…

emminizer:
… iterator invalidation from GeoTransform.

### DIFF
--- a/src/osgEarth/AnnotationNode
+++ b/src/osgEarth/AnnotationNode
@@ -27,7 +27,6 @@
 #include <osgEarth/DepthOffset>
 #include <osgEarth/MapNodeObserver>
 #include <osgEarth/Horizon>
-#include <osg/Observer>
 
 #define META_AnnotationNode(library,name) \
     META_Node(library,name)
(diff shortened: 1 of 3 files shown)


emminizer:
The issue is that the values where Observer were attached were already in an `observer_ptr`.  In `AnnotationNode`'s case, there's no benefit that I can see to having it be an `Observer`, because the set-to-`NULL` operation in the callback is already happening anyways, and nothing "extra" happens.

In the case of `GeoTransform`, extra stuff is happening on delete.  There are two problems with the `objectDeleted` implementation here though:

1. There's a comparison against the `observer_ptr` in the first line.  But that `observer_ptr` might (or might not) have gotten its `objectDeleted` call, and may already be `NULL`.  So it's not a great idea to do the comparison here because it might erroneously fail due to order of operations in OSG C++ code.  To fix, we don't compare, since it's only called when the object being observed is deleted and we can only ever observe the item that is `_terrain`.  So the check is superfluous.  I added checks to `setTerrain()` to make sure the checks in `objectDeleted` are superfluous.
2. Setting the `observer_ptr` to NULL will cause it to remove itself from a list of `Observers` that is currently being iterated.  This leads to iterator invalidation.  Instead, don't touch `_terrain` and let `observer_ptr` do its job, preventing iterator invalidation, which led to crashes.

PR #1520, opened 2020-05-15
FeatureTests: Fix ambiguous type error on gcc.

emminizer:


emminizer:
Tested by running on Windows MSVC 2019 and Linux gcc 8.3.1.  In both cases, I got failures on the test in something in Cache and Image Layer, but this Feature part works.  I did not investigate the Image Layer and Cache test failures.

jasonbeverage:
Thanks for the fix.  The cache tests will only work if you actually set the OSGEARTH_CACHE_PATH as an env var before hand.

Issue #1526, opened 2020-05-31
ImageControl not displaying image on osgearth 3.0


MCA4213:
Hi, I have noticed that ImageControl is displaying a white rectangle in the 3.0 version (no image).
Comparing the Controls.cpp file with older version 2.10 (that works) show that removing "g->setStateSet(getGeomStateSet());" in "void ImageControl::draw( const ControlContext& cx )" function **will fix the problem**.

gwaldron:
If you have a fix please submit a PR, thx

PR #1534, opened 2020-06-15
Fix issue #1526: ImageControl not show image, only white rectangle

MCA4213:
Hi, I have noticed that ImageControl is displaying a white rectangle in the 3.0 version (no image).
Comparing the Controls.cpp file with older version 2.10 (that works) show that removing "g->setStateSet(getGeomStateSet());" in "void ImageControl::draw( const ControlContext& cx )" function will fix the problem.


MCA4213:
Hi, sorry this is the first time that I make a pull request.
The fix is the first commit only!
Fix issue #1526: ImageControl not show image,

gwaldron:
Please create a new PR with just the one commit you intend, thanks.

gwaldron:
Fixed

PR #1561, opened 2020-08-11
Threading: SetThreadName now checks _WIN32_WINNT target version. Adds…

emminizer:
… path to support Win7 again.

### DIFF
--- a/src/osgEarth/Threading.cpp
+++ b/src/osgEarth/Threading.cpp
@@ -504,7 +504,7 @@ ThreadPool::get(const osgDB::Options* options)
 void
 osgEarth::Threading::setThreadName(const std::string& name)
 {
-#if (defined _WIN32 && defined _WIN32_WINNT_WIN10) || (defined __CYGWIN__)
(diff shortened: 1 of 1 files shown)


emminizer:
Some more details beyond the commit message:
We are building osgEarth on MSVC 2019, which has a newer Windows SDK by default than some of the older MSVC versions.  As such, the Win10 preprocessor define is defined.  I can, on a Windows 7 (I know) machine build the software, but then can't run it on the same machine.

Adding this check by itself does not cause the code to work, because if undefined, `_WIN32_WINNT` will default to `_WIN32_WINNT_WIN10`.  But this gives us an opportunity to set the target `_WIN32_WINNT` in our CMake configuration, allowing us to target older versions than Windows 10.

Without this change (and a define in CMake command line / `CMAKE_CXX_FLAGS`), osgEarth builds on MSVC 2019 in such a way that it will only run on Windows 10 systems.  With this change, that's still the same default, but you can specify CXX flags to override that behavior.

PR #1562, opened 2020-08-12
FileUtils: getTempName() adds thread ID, making it more robust while threading

emminizer:
I am using GDALDEM driver, which is getting lots of errors on the console about various TIF files already existing or not existing.  This has led to crashes at runtime.  This change works around that problem by encoding the thread ID into the temporary filename, which helps prevent multi-threaded failures.  As a result, GDALDEM no longer crashes sporadically and I haven't seen any filename errors yet.

### DIFF
--- a/src/osgEarth/FileUtils.cpp
+++ b/src/osgEarth/FileUtils.cpp
@@ -331,7 +331,7 @@ osgEarth::Util::getTempName(const std::string& prefix, const std::string& suffix
     while (true)
     {
         std::stringstream ss;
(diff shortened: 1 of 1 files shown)


jasonbeverage:
Hey Dan can you do me a favor and change the OpenThreads::Thread::CurrentThreadId() call to osgEarth::Threading::getCurrentThreadId() and I'll merge your PR?  OpenThreads::Thread::CurrentThreadId() has issues with threads that aren't created with openthreads.

emminizer:
Sure thing.  I wasn't aware of the issue.  Yes, that works well on our side too.  Updated and good to go.

PR #1568, opened 2020-08-26
Changed include references from osgEarth/ThreadingUtils

nmielcarek:
to osgEarth/Threading due to ThreadingUtils no longer existing in version 3.0. osgEarth/Threading seems to be included in several lower level includes so its possible these lines can be removed completely.


GouShiQuan:
unsubscribe



At 2020-11-17 02:07:11, "Glenn" <notifications@github.com> wrote:

Merged #1568 into master.

PR #1578, opened 2020-09-15
Exposing struct "CameraData" to dll

Adellbengbeng:
Please append "OSGEARTH_EXPORT" to the declaration of struct "CameraData". Otherwise,  "CameraData::~CameraData()" won't be found when building "osgearth_geodetic_graticule.exe".

### DIFF
--- a/src/osgEarth/GeodeticGraticule
+++ b/src/osgEarth/GeodeticGraticule
@@ -141,7 +141,7 @@ namespace osgEarth { namespace Util
         osg::Vec2f _centerOffset;
         std::vector< double > _resolutions;
 
-        struct CameraData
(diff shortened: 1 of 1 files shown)


HyperEye:
I had to apply the same fix when attempting to build with mingw-w64.

JeffSWhite:
I had the same issue that I described here:

http://forum.osgearth.org/GeodeticGraticule-linker-issue-with-MinGW-td7593376.html

The proposed fix took care of the issue.

GouShiQuan:
unsubscribe



在 2020-10-30 04:55:47，"Glenn" <notifications@github.com> 写道：

Merged #1578 into master.

PR #1579, opened 2020-09-18
Added setting of VertexArrayState VBO usage flag to ensure VBO is bound if necessary

JeffSWhite:
Added setting of VBO usage flag in VertexArrayState prior to drawing in SharedGeometry::drawImplementation() because sometimes (at least on Windows) the OSG StatsHandler would change this VBO usage flag to false prior to SharedGeometry::drawImplementation() being called.  This would cause the VBO to not be bound when glDrawElements was called and resulted in crashes.


gwaldron:
Thanks Jeff!

PR #1588, opened 2020-10-13
fix 3dtiles refine bug

showintime:
I found that if the Tile corresponding to ThreeDTileNode does not have a `refine` member, osgEarth cannot handle the `refine `method correctly.
If the Tile corresponding to ThreeDTileNode does not have a `refine` member, according to the 3dtiles specification, we should inherit the `refine` method of the parent Tile, and this process should be at the beginning of the ThreeDTileNode, instead of inheriting the `refine` method of the parent Tile after the construction is completed, because at this time The `refine` of the inherited parent Tile is likely to be REFINE_ADD initialized by default.


showintime:
Since the `refine` of Tile is inherited from the parent Tile, we need to traverse down from the root Tile to get the correct `refine` of the Tile. If the parent Tile is not initialized in the constructor of ThreeDTileNode, the refine inheritance of setParentTile to Tile will be From the leaf node to the root node, this will get the wrong `refine`, unless we traverse twice, the first time from the root tile down to initialize ThreeDTileNode, the second time from the root tile down to set up `refine`. do you mean this?

jasonbeverage:
If you remove the initialization if the _refine then it will be under, so when setParentTile is called the node should inherit the refine of it's direct parent, which should have been either set explicitly in the tileset.json or inherited by it's parent.  

Do you have a 3d tiles dataset that shows the issue and I can look into it here?

showintime:
[Melbourne Photogrammetry](https://cesium.com/ion/assets/69380)
see [earth file](https://cesium.com/blog/2020/04/02/osgearth-supports-cesium-ion-assets/)

showintime:
cesium result
屏幕截图(42)
osgearth result
屏幕截图(41)

jasonbeverage:
Great, I pushed a change in ce5fdfe5bd19230c645f0e236543aaa1f78eb37b that should fix this issue.  Thanks!

PR #1606, opened 2020-11-23
Dev/showintime/stack initial 3dtile node

showintime:
Ensure that the refine of each Tile is initialized correctly.


jasonbeverage:
Thanks for this, the one liner I pushed seemed to work just fine!

PR #1608, opened 2020-11-28
3.1: github actions for windows, linux and macos

remoe:
Add github actions to build osgEarth 3.1 on windows, linux and macos. "vcpkg" only works with windows. using xcode 12 on macos. I've tested it on my osgEarth 3.1 branch. All builds successed, but one need to test/adapt it for osgEarth repository.

All action have "workflow_dispatch". To build it on push, one can change this to "push". 
Attention: The "windows action" use "vcpkg" and creates "nuget" packages on your current github account!

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -1,0 +1,70 @@
+name: linux-x64 
+
+on: [workflow_dispatch] # workflow_dispatch push
+
+env:             
+  # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
+  BUILD_TYPE: Release
+
+jobs:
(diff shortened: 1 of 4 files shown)


jasonbeverage:
Awesome Remo, give us some time to look at this and we'll probably get it merged.

remoe:
@gwaldron @jasonbeverage the vcpkg - cache is bigger then 5GB in last osgEarth/windows-action-build. I don't think vcpkg need so much storage to build. One need to optimize the folders to exclude here:

https://github.com/gwaldron/osgearth/blob/gh-actions/.github/workflows/windows.yml#L37

PR #1612, opened 2020-12-02
SilverLining: Fix older API compatibility

emminizer:
Removed the ifdef from header, moved to .cpp, to fix API compatibility issue.


emminizer:
Fixed by https://github.com/gwaldron/osgearth/commit/40d8dc29373bf49027ad8040ac13972f4462c82d

PR #1626, opened 2021-01-21
Threading: Fix MSVC 2015 build errors

emminizer:
MSVC 2015 fails on latest master unless typename is added.  Tested change on MSVC 2015 and on g++ 8.3.


emminizer:
A better approach:
* Remove typename
* Remove extra qualification "Job::" in these lines, matching the style from other dispatch function above

Closing out PR as there is a better solution.

PR #1647, opened 2021-02-12
fix: MacOSX build issue with ExampleResources

remoe:
It doesn't have a default template for this enum.


remoe:
The MacOSX github action fails only because it doesn't checkout the branch for this patch.

PR #1649, opened 2021-02-17
Utils: Fix compile error with incorrect variable name.

emminizer:
Did a quick look at the code after seeing compile failure on gcc.  I think this is what was intended.


emminizer:
https://github.com/gwaldron/osgearth/commit/fdd712c2c93a5a063521b327240634445767c972 fixes this instead by renaming the parameter.  One more fix needed still after that commit to fix the build error.  Closing this PR out since that change breaks the merge of this.

PR #1652, opened 2021-02-22
Rex Loader: Fix gcc compile issue with reference to temporary.

emminizer:
Fixes gcc 8.3.1


emminizer:
OBE, `clear()` now just assigns empty vectors.

PR #1660, opened 2021-03-13
Change webp -> libwebp in build.md

ellerychan:
The instructions say to install the optional dependency `webp:x64-windows` which throws the error:

```
Error: while loading web:
The port directory (C:\src\vcpkg\ports\webp) does not exist
Error: failed to load port from C:\src\vcpkg\ports\webp
Note: Updating vcpkg by rerunning bootstrap-vcpkg may resolve this failure.
```

Changing the package name to `libwebp:x64-windows` removes the error.  Was the package name changed recently?  I'm not certain that this is the correct fix, so someone in-the-know should verify that it is.

### DIFF
--- a/docs/source/build.md
+++ b/docs/source/build.md
@@ -21,7 +21,7 @@ vcpkg install osg:x64-windows gdal:x64-windows curl:x64-windows
 For full functionality, you can install optional dependences as well:
 
 ```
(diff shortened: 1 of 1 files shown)


gwaldron:
Thanks for spotting the error. Merged.

PR #1666, opened 2021-04-07
Registry: Revert 957f487d change, now registry ref_ptr is global inst…

emminizer:
…ead of local to instance().

Fixes order of destruction issue when using e.g. ImageOverlay, where _program destruction happens after ~Registry, and ImageOverlay's VirtualProgram accesses a stale ProgramRepo reference, leading to a crash.

### DIFF
--- a/src/osgEarth/Registry.cpp
+++ b/src/osgEarth/Registry.cpp
@@ -211,27 +211,43 @@ Registry::~Registry()
     CPLPopErrorHandler();
 }
 
+static osg::ref_ptr<Registry> s_registry = NULL;
+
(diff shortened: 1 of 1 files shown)


emminizer:
This is not a perfect fix.  It works around the problem by not allowing instance() to return a stale pointer.  This prevents crash on exit.  My use case is a standalone application running a test where the application creates an ``ImageOverlay`` instance.  Its static ``_program`` gets destroyed after the Registry is destroyed, which leads to the crash I'm seeing.

The crash was introduced in 957f487d, but we didn't update past that revision internally until yesterday.  Sorry for noticing the problem late.

Issue #1691, opened 2021-05-08
Multiple problems with POINTSET geometries


MCA4213:
Hello, I have noticied at two problems with POINTSET geometrie:
1: The first one is related to the AnotationNode, it draw only one point at the centroid, even if a have 10!!!
I am using this code:
```
 osgEarth::Style style;
       style.getOrCreate<IconSymbol>()->url() = osgEarth::StringExpression("../data/images/pin.png");
       style.getOrCreate<IconSymbol>()->placement() = osgEarth::IconSymbol::PLACEMENT_VERTEX;
       osg::ref_ptr <osgEarth::Feature> feature = new osgEarth::Feature(observer->geom, srsWgs);
       observerZone->setFeature(feature);
       observerZone->setStyle(style);
       observerZone->dirty();
```
If I change to modelSymbol, it will display all the points correctly. **(I don't knew how to fix this first problem)**

2: All GEOS operations with POINTSET fails. This because `GEOS::importGeometry` is failling, and this is caused by the manner that you create the geometry.
At `GEOSGeometry*  import(GEOSContextHandle_t handle, const Geometry* input) ` in **GEOS.cpp**, you are creating a pointset with the same manner as you do for point using GEOSGeom_createPoint_r, but this fonction can create only one point and fail if you give more than one!
So as a quick solution, you can create a multiple points geometries, than add them to a collection, and this will make it work.
 ```
case Geometry::TYPE_POINT:
                seq = vec3dArray2CoordSeq(handle, input, false);
                if (seq) output = GEOSGeom_createPoint_r(handle, seq);
                break;
```

```
case Geometry::TYPE_POINTSET:
                geomsList=new GEOSGeom[input->size()];
                for (int i = 0; i < input->size(); i++)
                {
                    seq = GEOSCoordSeq_create_r(handle, 1, 3);
                    {
                        GEOSCoordSeq_setX_r(handle, seq, 0, input->at(i).x());
                        GEOSCoordSeq_setY_r(handle, seq, 0, input->at(i).y());
                        GEOSCoordSeq_setZ_r(handle, seq, 0, input->at(i).z());
                    }
                    geomsList[i] = GEOSGeom_createPoint_r(handle, seq);
                }
                output = GEOSGeom_createCollection_r(handle, GEOS_MULTIPOINT, geomsList, input->size());
                delete geomsList;
                break;
```
**(This could be a solution for problem 2)**

[...]

PR #1707, opened 2021-05-21
fix second GEOS issue with Pointset ( issue #1691 )

MCA4213:
This is a fix of how to create a pointset using GEOS.

Problem: GEOS::importGeometry is failling when creating a pointset  using GEOSGeom_createPoint_r. this fonction can create only one point and fail if you give more than one!

### DIFF
--- a/src/osgEarth/GEOS.cpp
+++ b/src/osgEarth/GEOS.cpp
@@ -95,16 +95,34 @@ namespace
         {
             // any other type will at least contain points:
             GEOSCoordSequence* seq = 0;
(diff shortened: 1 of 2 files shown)


gwaldron:
Let me know if you plan to address the other POINTSET issue; I recall there were two. If so we will try to get it into version 3.2. Thanks.

MCA4213:
Hi, I don't think that I will be able to resolve the first issue, because I don't know where to look, if I can get some hints about the classes responsibles of this, perhaps I can help. Sorry.

PR #1763, opened 2021-07-23
Github Action Windows: change cache key

remoe:
One cannot delete the cache. Try to change the key.


remoe:
@gwaldron Windows build is fixed.

Issue #1703, opened 2021-05-20
OSX: osgEarth master (3.2) not show the globe


remoe:
Run osgearth_viewer with 9c4289e6b6c4af53c3960f5dc23bb28115d6d23e does not show the globe (readymap.earth):

<img width="547" alt="image" src="https://user-images.githubusercontent.com/804941/119005933-332dc980-b990-11eb-8d3e-8a1b9e68cd0e.png">

Log:

```
2021-05-20 17:20:07.475127+0200 osgearth_viewerd[24366:936398] Metal API Validation Enabled
2021-05-20 17:20:07.516349+0200 osgearth_viewerd[24366:936927] flock failed to lock maps file: errno = 35
2021-05-20 17:20:07.516827+0200 osgearth_viewerd[24366:936927] flock failed to lock maps file: errno = 35
[osgEarth]  [Capabilities] Capabilities: 
[osgEarth]  [Capabilities]   osgEarth Version:  3.1.0 build 110
[osgEarth]  [Capabilities]   OSG Version:       3.7.0
[osgEarth]  [Capabilities]   GDAL Version:      3.2.1
[osgEarth]  [Capabilities]   GEOS Version:      3.6.2
[osgEarth]  [Capabilities]   GPU Vendor:        ATI Technologies Inc.
[osgEarth]  [Capabilities]   GPU Renderer:      AMD Radeon Pro 560 OpenGL Engine
[osgEarth]  [Capabilities]   GL/Driver Version: 4.1 ATI-3.10.19
[osgEarth]  [Capabilities]   GL Core Profile:   yes
void StateSet::setGlobalDefaults() ShaderPipeline disabled.
[osgEarth]  [EarthSerializer2] Loaded extension lib "osgPlugins-3.7.0/osgdb_osgearth_viewpointsd.so"
[osgEarth]  [Map] Map profile is: [srs=WGS84, min=-180,-90 max=180,90 lod0=2,1 vdatum=geodetic]
[osgEarth]  [Map] [no cache]
[osgEarth]  [ViewpointsExtension] Read 19 viewpoints
[osgEarth]  [TileLayer] Layer "ReadyMap 15m Imagery" [srs=WGS84, min=-180,-90 max=180,90 lod0=2,1 vdatum=geodetic] [no cache]
[...]

PR #1769, opened 2021-08-16
Rex SharedGeometry: Refactor draw primitives implementation to fix ATI rendering issues with older drivers.

emminizer:


gwaldron:
Merging prior to 3rd party app testing. If there's an issue we may revisit

remoe:
@emminizer , how was the behavior of this rendering issues?

emminizer:
@remoe We have been getting reports for the last several months from users who can no longer view the globe. These were all users with older cards (e.g. Matrox C420 with 22.19.693.256 driver, AMD Radeon HD 7730M) that were experiencing this issue. Drivers were as upgraded as they could get (newer drivers don't support older card).

The symptom manifested essentially as a blank, black globe. With the simple sky model enabled, we could see the lighting along the edge of the globe, but no texture data. We could add things like line features onto the globe and they were drawn correctly, it was just the underlying globe imagery data that wasn't rendering. This used to work on software releases 9+ months back.

We were able to generate API Trace output from a failing laptop. When running the replay through API Trace, the display was incorrect when run on the AMD Radeon HD 7730M. But when we ran the exact same trace file through API Trace on systems with other cards, everything was looking fine.

Current belief is that there's something funky with the VAO state and EBO that isn't being captured by these older cards, forcing a requirement to generate and rebind it each time.

remoe:
@emminizer thanks for this detailed explanation! We have similarly (VAO) issues with MacOSX:
https://github.com/gwaldron/osgearth/issues/1703

emminizer:
Oh! Yes that very likely could be the same issue we were seeing. I had not seen that issue previously.

PR #1775, opened 2021-08-23
Valgrind fixes: ~Map is now protected because it is a ref_ptr; unused…

emminizer:
… _ElevationPool removed; fixed leak of ElevationPool::_index.


emminizer:
Just realized commit comment is wrong, `~ElevationPool` was made protected, not `~Map`. Sorry about that.
The extra _ElevationPool was found by making its destructor of Elevation Pool protected. It looked to be completely unused.
Valgrind was complaining about the leaking _index on exit.

PR #1787, opened 2021-09-10
ScreenSpaceLayout: Enable state now correctly updates for callouts, w…

emminizer:
…hen text priority or declutter state changes while label is valid but not drawn.


emminizer:
We have a report where a LabelNode is incorrectly showing the callout for the screen space layout, when the user thought the declutter was disabled for this particular label (via `TextSymbol::declutter()` and via `LabelNode->setPriority()`).

I was finally able to reproduce the bad behavior by rendering the label, seeing the callout, then changing our scenario such that the label is not visible. For our time-based application, this simply meant looping time around to before the entity creation. The label was still valid, just not visible. At this time, the priority and declutter flags update (verified with cout). And simultaneously, the label is turned back on.

I was able to fix this by checking the fields in the `layoutData` more regularly, allowing them to turn on or off the `Element::_declutter` flag dynamically. After this change I no longer see bad behavior on my end.

PR #1795, opened 2021-09-14
Simple Sky: Moon and Sun no longer render in ortho views.

emminizer:


emminizer:
We came across an issue where we were using overhead mode in our tool, which renders everything with an ortho view. The sun and moon were gigantic, and were rendering improperly.

This is a similar fix to what was applied to atmosphere in https://github.com/gwaldron/osgearth/commit/79cbedfa44

PR #1851, opened 2021-11-05
Rework Threading internal queue

Davidbrcz:
This commit rework Threading internal queue

- It replaces the vector + partial sorting with a std::priority_queue
- In introduces a proper abstraction for the internal queue of job
SynchronizedPriorityQueuedJob. By doing so,
	- it removes a datarace that was happening on _done
	- it ensures that the variables used in the condition_variable
	are always modified while holding the lock, preventing improper use
	of block with predicate (as there is small window between the predicate
	call and the lock which cam lead to lost notifications)
- It refactors the code with this new SynchronizedPriorityQueuedJob.


src/osgEarth/Threading.cpp
```diff
@@ -743,57 +739,28 @@ JobArena::runJobs()
     // cap the number of jobs to run (applies to TRAVERSAL modes only)
     int jobsLeftToRun = INT_MAX;
 
-    while (!_done)
+    while (_queue.running())
     {
-        QueuedJob next;
-
-        bool have_next = false;
+        if(_type == update_traversal)
```
Makamitsu, line 744: update_traversal should be uppercase

Davidbrcz: I don't know how that happened....

Davidbrcz:
I've
- rebased the PR, fixed the compilation issue 
- made shutdownAll publc and remove its registration with atexit
- changed startThreads to make the thread creation much less dependent on the actual thread scheduling.

PR #1925, opened 2022-01-10
fix: gha - windows build cache failure

remoe:
Fix gha cache failure:

https://github.com/gwaldron/osgearth/runs/4741870605?check_suite_focus=true

gha-cache-failed


remoe:
@gwaldron , this should fix the current Github CI Windows build issue.

PR #1961, opened 2022-02-01
GLUtils: Remove extra class specification in front of function, fixing some builds.

emminizer:
Building against master osgEarth I'm seeing errors due to this, on MSVC 2019. Might be related to /sdl flag, did not investigate though.


gwaldron:
Thanks Dan, I made this fix before seeing your PR :)

PR #1962, opened 2022-02-01
glEnableClientState: Getting pointer from osg::setGLExtensionFuncPtr() for use in GL Core profile.

emminizer:
The glcorearb.h does not define `glEnableClientState()`. When compiling in core profile, I cannot use `glEnableClientState()` directly. This patch allows for use of `glEnableClientState()` in core profile. I tested that it returns a valid non-NULL pointer on my NVidia card. I did not test compatibility mode because I do not have a compatibility mode build of OSG/osgEarth.


gwaldron:
Dan, I made these changes manually. Thanks for the PR.

PR #1969, opened 2022-02-04
FeatureSource: Avoid null dereference in rare case where feature prof…

emminizer:
…ile is not set, such as in some cases when loading a VTPK file.


emminizer:
Got into a crash condition on start-up in our application when trying to initialize with a VTPK-related earth file. This was the only layer in the earth file, and it did not open up properly due to invalid paths.

PR #2014, opened 2022-03-11
TextureArena: Fix crash on exit from accessing invalid array index.

emminizer:
Crashes consistently on exit on Linux. On Windows I'm seeing odd behavior but it's not crashing.


emminizer:
I am no longer seeing the crash on exit after https://github.com/gwaldron/osgearth/commit/c9abe9f4672ce9575bd88bd0e4b3e1ed6df753e8

gwaldron:
Fixed a different way. Thanks.

PR #2047, opened 2022-04-21
GLUtils: Fix create() call with invalid parameter, fixing gcc build

emminizer:


gwaldron:
Fixed independently. Thanks.

PR #2077, opened 2022-05-27
ImageUtils::ColorWriter float/double mixup

mk401:
Fix for issues found stemming from [f551c75](https://github.com/gwaldron/osgearth/commit/f551c750acbc740fa83c3bbdc037a6705cea0dca?diff=split). `GLTypeTraits<>::scale()` all return double. Float conversion issues are causing artifacts near edges, preventing some pixel-testing from working, which we use for transparent image guessing.

### DIFF
--- a/src/osgEarth/ImageUtils.cpp
+++ b/src/osgEarth/ImageUtils.cpp
@@ -1757,7 +1757,7 @@ namespace
     {
         static void write(const ImageUtils::PixelWriter* iw, const osg::Vec4f& c, int s, int t, int r, int m )
         {
-            float scale = GLTypeTraits<T>::scale(iw->_normalized);
+            double scale = GLTypeTraits<T>::scale(iw->_normalized);
             T* ptr = (T*)iw->data(s, t, r, m);
             *ptr++ = (T)( c.r() / scale );
             *ptr   = (T)( c.a() / scale );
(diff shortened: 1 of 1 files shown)


jasonbeverage:
I'm surprised this is an issue, especially only on writing as I see the read is still using floats.  Can you give an example of what kind of issues you're seeing or provide a bit of example code showing the issue?

mk401:
I think it's only an issue on writing because the `ColorReaders` cast the results of the pointer increments to float before performing an operation, whereas the `ColorWriters` don't.

As an example, the image we're sending through this code via `osgEarth::ImageUtils::convertToRGBA8()` is white .bmp image with a small blue arrow in the middle. We convert it to RGBA format, then run a transparent color guess algorithm on the image (adapted from FOX code) to determine what color we should be replacing with transparent.

Before the change in the commit referenced above, `ColorWriter` for `GL_RGBA` correctly converted the pixels to 255. After the change, the RGBA values are 254.

The image resulting from `osgEarth::ImageUtils::convertToRGBA8()` returns true from `osg::Image::isImageTranslucent()`, when it didn't before, which we use as an early-out on our own code when guessing a color to replace with transparent. Hope this all makes sense.

jasonbeverage:
Thanks for the explanation, makes sense to me and thanks for the fix!

PR #2082, opened 2022-06-13
ArcGISTilePackage: Remove extraneous slash that caused ArcGIS Tile Package to fail under Linux

emminizer:
`osgEarth::getFullPath()` behaves differently when absolute paths are passed into the second parameter. This removes the leading slash, which is used to indicate absolute path on Linux, fixing display of ArcGIS Tile Packages on Linux.


gwaldron:
Thanks!

PR #2094, opened 2022-07-18
Fix target names in osgEarthConfig.cmake.in

cbrl:
The osgEarthConfig.cmake.in file does not create an `osgEarth` target, which causes configuration to fail in CMake upon calling `set_target_properties(osgearth ...)` on a nonexistent target here: https://github.com/gwaldron/osgearth/blob/218e23fb7fee6443dcfeec3d60f0d2c79dbbedde/osgEarthConfig.cmake.in#L41

---

In addition, I believe the variable and target names are not capitalized correctly. As defined in src/osgEarth/CMakeLists.txt, the library name is "osgEarth".
https://github.com/gwaldron/osgearth/blob/218e23fb7fee6443dcfeec3d60f0d2c79dbbedde/src/osgEarth/CMakeLists.txt#L53

However, all variable and target names in osgEarthConfig.cmake.in use the lowercase "osgearth", which can result in unexpected behavior or configuration errors. With standard variables such as `<library_name>_INCLUDE_DIRS`, `<library_name>` is expected to exactly match the spelling of the target.

---

The final issue I encountered is the capitalization of the path provided to `find_path()`. This can cause problems finding the path to osgEarth. I've observed that it does not find the correct path on a system running Ubuntu 20.04.
https://github.com/gwaldron/osgearth/blob/218e23fb7fee6443dcfeec3d60f0d2c79dbbedde/osgEarthConfig.cmake.in#L16

### DIFF
--- a/osgEarthConfig.cmake.in
+++ b/osgEarthConfig.cmake.in
@@ -5,53 +5,54 @@ set(OSGEARTH_VERSION 3.1.0)
 set(XPREFIX OSGEARTH)
 
 
-
-if (osgearth_USE_STATIC_LIBS)
+if (osgEarth_USE_STATIC_LIBS)
     set(XPREFIX ${XPREFIX}_STATIC)
 endif()
 
-set(osgearth_DEFINITIONS ${${XPREFIX}_CFLAGS})
+set(osgEarth_DEFINITIONS ${${XPREFIX}_CFLAGS})
 
-find_path(osgearth_INCLUDE_DIR
-    NAMES OSGEARTH/RTREE.H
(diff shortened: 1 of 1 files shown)


plevy:
Thanks Connor, the initial was just a first try to get find_package() support while updating vcpkg port for 3.3 at the same time.  Maybe not so great support on linux so thank you for making it better at multiple platforms

cbrl:
No problem. Windows seems to be more lenient on the rules sometimes. I don't experience any issues with the existing package configuration on that platform.

PR #2129, opened 2022-08-26
ImageUtils nnUVtoST: Replace IN and OUT type with T_IN and T_OUT to a…

emminizer:
…void define conflicts with Windows SDK headers wincrypt.h, minwindef.h, bcrypt.h, and rpcdce.h.


emminizer:
Excerpt from `wincrypt.h`:

```
#ifndef IN
#define IN
#endif

#ifndef OUT
#define OUT
#endif
```

Was getting compile errors in user level code. osgEarth compiles fine, but including osgEarth with certain other libraries led to compilation errors in this code.

gwaldron:
Thanks, already fixed by changing the template param names.

emminizer:
Thanks. I thought I was up to date when I posted this but I see your template parameter name change now. Sorry for the noise in that case.

PR #2132, opened 2022-09-02
ExampleResources.cpp gets -bigobj on MSVC 2019 and older, fixing debug compilation issue

emminizer:


jasonbeverage:
Thanks Dan, we might look at enabling this globally or slimming down ExampleResources here at some point.

Issue #2138, opened 2022-09-16
[bug] Windows build error:  C1128: number of sections exceeded object file format limit


remoe:
Build error with current master f6b92f66ed314b85f2deb6cadd907e0df4426052 :

```log
osgearth.git\src\osgEarth\MapboxGLImageLayer.cpp : fatal error C1128: number of sections exceeded object file format limit: compile with /bigobj [osgearth.git\_build\src\osgEarth\osgEarth.vcxproj]
osgearth.git\src\osgEarth\ObjectIDPicker.cpp : fatal error C1128: number of sections exceeded object file format limit: compile with /bigobj [osgearth.git\_build\src\osgEarth\osgEarth.vcxproj]
osgearth.git\src\osgEarth\PowerlineLayer.cpp : fatal error C1128: number of sections exceeded object file format limit: compile with /bigobj [osgearth.git\_build\src\osgEarth\osgEarth.vcxproj]
cl : Command line error D8040: error creating or communicating with child process [osgearth.git\_build\src\osgEarth\osgEarth.vcxproj]
```

platform: windows
compiler: vs2017 / MSVC 19.16.27045.0

PR #2139, opened 2022-09-16
Fixing VS2017 debug compilation issue

remoe:
It needs more files to use ```/bigobj``` on debug build then it was noted on #2138

### DIFF
--- a/src/osgEarth/CMakeLists.txt
+++ b/src/osgEarth/CMakeLists.txt
@@ -848,10 +848,26 @@ set(TARGET_SRC
     ${SHADERS_CPP}
 )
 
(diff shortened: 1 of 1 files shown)


emminizer:
I agree. I started to see debug issues on MSVC 2019 too (but not 2022). It might be worthwhile to just enable `/bigobj` on all debug builds on MSVC.

jasonbeverage:
Hi Remo!  Good to hear from you, it's been awhile.

I pushed a change implementing Dan's suggestion of just enabling /bigobj globally on Debug builds instead of trying to do it per file.  At some point we might want to just enable it globally no matter what.  Based on this discussion https://developercommunity.visualstudio.com/t/Enable-bigobj-by-default/1031214 it seem like it's more about backwards compatibility with older versions of VS than any kind of performance implication.

PR #2170, opened 2022-10-27
ImGui tools includes and namespace fixes

mk401:
Fix minor includes and namespace issues in ImGui and ImGui tools


mk401:
some context: We are expanding our usage of ImGui in our SDK examples to include the osgEarth ImGui tools you've written (e.g. SceneGraphGUI), and a few files needed updates to build on their own, outside of the osgEarth examples and ImGuiApp files they're used in conjunction with on your end.

jasonbeverage:
Thanks for the PR!  I'm glad the ImGUI stuff is proving useful to others.  I originally just hacked it together to write some tests and now I can't live without it :)

PR #2192, opened 2022-12-12
URI: Fixed gcc compile bug with temporary reference.

emminizer:
Fixes compile issue on Linux gcc 8.3.1 on master


emminizer:
I see there's an error on the build. I don't see that error on my Linux build with gcc 8.3. Applied fix regardless, appears to now pass GitHub Linux/MacOS builds.

PR #2216, opened 2023-02-01
EarthManipulator: Decouple Pen/Touch Drag from Left Mouse Drag

emminizer:
Touch improvements on `EarthManipulator`, including:
* `Settings` gets new `bindTouchDrag()`, defaulting to `ACTION_PAN`. New `EVENT_TOUCH_DRAG` to support this.
* Single touch now uses the new `EVENT_TOUCH_DRAG` instead of `EVENT_MOUSE_DRAG` from left mouse
* Multi-touch now works with 2 or more fingers, rather than just 2 fingers, ensuring that single touch/drag only ever gets a single finger.

This is most useful in cases where the `EarthManipulator` settings for left click/drag are different from what you'd want in the touch/drag case. This is the case with SIMDIS SDK, where we change the default left drag operation but doing so makes touch operations basically unusable.

This does introduce a new enum and public method, and I recommend an osgEarth API version update if/when this is merged in (I can add it to PR if you want).

### DIFF
--- a/src/osgEarth/EarthManipulator
+++ b/src/osgEarth/EarthManipulator
@@ -79,7 +79,8 @@ namespace osgEarth { namespace Util
             EVENT_MOUSE_CLICK        = osgGA::GUIEventAdapter::USER << 1,
             EVENT_MULTI_DRAG         = osgGA::GUIEventAdapter::USER << 2,   // drag with 2 fingers
             EVENT_MULTI_PINCH        = osgGA::GUIEventAdapter::USER << 3,   // pinch with 2 fingers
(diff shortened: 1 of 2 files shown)


emminizer:
I tested `osgearth_imgui` and `osgearth_viewer` before and after these changes, on a touch screen and when using the mouse. I saw no difference in observable functionality from user's perspective. A dev would notice a difference in their own software on touch displays, only if they create their own blank `EarthManipulator::Settings` rather than using the default values.

jasonbeverage:
Thanks for this Dan!

PR #2217, opened 2023-02-03
ShaderUtils: _uniformAlt now properly added into state set

emminizer:
Uncommented code block that was commented out in 86a72d202. By commenting this out, SIMDIS SDK's capability to accept multiple projectors on a single entity is broken. By uncommenting it, the functionality is fixed again. It looks like this might have been commented out temporarily / accidentally in that changeset to test the impacts of other fixes in the code.


gwaldron:
Oops. Thanks Dan.

PR #2259, opened 2023-04-07
GeoPoint: Avoid null access on SRS

emminizer:
Avoids a null access on SRS that can be triggered with the following code block:

```
osgEarth::Style style;
auto label = new osgEarth::LabelNode("crash", style);
```

Introduced in https://github.com/gwaldron/osgearth/commit/9454d221


gwaldron:
Thanks Dan!

PR #2275, opened 2023-04-21
fix: macosx and windows ci/cd build issue

remoe:
This should fix the build issues with Windows and MacOSX


jasonbeverage:
Thanks Remo!

remoe:
The windows build had a 6h default timeout. The next commit should need fewer time.

PR #2289, opened 2023-05-17
AnnotationsGUI: Fix namespace errors

emminizer:
Getting build errors when including this in other code. Appears to be missing `osgEarth::` namespace, due to being in anon namespace without a `using` statement.


gwaldron:
Thanks Dan.

PR #2331, opened 2023-09-07
Fix for loading TFS layers based on min/max levels available in the s…

dsallen:
…ource data without explicit levels set in the TFSLayer configuration in osgEarth.

You can reproduce the issue by loading the example feature_tfs


gwaldron:
Sorry, but I don't understand what this PR is trying to fix.
How do we reproduce the problem you're seeing? The example `feature_tfs_scripting.earth` is functioning properly.

PR #2370, opened 2023-10-31
Fixed link for webp, now using TARGET_LIBRARIES_VARS

emminizer:
WebP plug-in now works in both release mode and debug mode.

In 6ac221f the linking statements got updated dramatically. This caused a ripple effect such that webp plug-in would work in release mode MSVC 2022, but in debug mode MSVC 2022 it built and failed to run. This is because after that changeset, webp started to link to `osg.lib` in release mode AND debug mode. Before this changeset, it linked to `osgd.lib`.

This PR fixes the behavior by using the `TARGET_LIBRARIES_VARS` instead of editing `TARGET_COMMON_LIBRARIES`. This allows the CMake code to pick the debug libraries for OSG when configured with debug libraries. After this change, webp starts working again in debug mode.

This has no impact on release osgEarth, but fixes a critical break in debug osgEarth.


gwaldron:
Thanks as usual Dan ;)

PR #2387, opened 2023-11-24
Dockerfile upgrade

elbarto1980:
This is a docker upgrade, based on my current experience on OSG and osgearth.
It uses Ubuntu 20.04.6 LTS (Focal Fossa) as base image and installs both OpenScenegraph 3.6.5 and osgearth.
@jasonbeverage, feel free to modify it as you see fit.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+.git
+.github
+.dockerignore
+Dockerfile
+docker-compose.yml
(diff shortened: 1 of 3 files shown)


jasonbeverage:
Thanks for the PR!  I was mostly using this Dockerfile for doing some data processing with osgearth_conv.  Does this actually allow you to run osg and osgEarth within docker with nvidia acceleration?

jasonbeverage:
Awesome!  Have you tried this on Windows at all or will this only work on Linux?

elbarto1980:
I am sorry but I don't have a Windows pc to try it on. I switched to Linux 7 yrs ago and never turned back. 😅

Issue #2383, opened 2023-11-19
Error in the screen display in Arch Linux.


samtux:
I have managed to compile osgEarth v3.5 on Arch Linux using the following instructions https://gist.github.com/samtux/63cc5df9d43a5fbc944742547ed0957f

However, when I try to display a view using osg_viewer, the screen flickers on screen as in the following animation:

![osgearth4](https://github.com/gwaldron/osgearth/assets/545152/26b11f43-f6c1-423e-a054-e6c3a6ad8ddb)

In the console show: 
```
❯ osgearth_viewer /usr/share/osgearth/test/osm.earth
[osgearth warning] [Capabilities] Warning, OpenSceneGraph does not define OSG_GL3_AVAILABLE; the application may not function properly
```
[](url)
Is there any recommendation you can give me to solve this problem?

Platform: Arch Liux x86_64.

gwaldron:
Would you please post the output of `osgearth_viewer --caps` ?
And the version of mesa you are using?
Thanks.

emminizer:
This might be fixed after https://github.com/gwaldron/osgearth/commit/e3ece4eed78f

gwaldron:
Mesa driver ticket:
https://gitlab.freedesktop.org/mesa/mesa/-/issues/10482

gwaldron:
Fixed by the Mesa team
https://gitlab.freedesktop.org/mesa/mesa/-/commit/f45c9a38db78776551ddee28c31a3a20cce287b6

samtux:
Thnaks, It is already working properly.

PR #2401, opened 2023-12-11
Rex: Work-around for MESA 23.1.4 bug with texture coordinates

emminizer:
This addresses the texturing aspect of the Mesa 23.1.4 bug. Terrain morphing is still an issue that needs to be addressed, but is significantly less visible than the texturing aspect that this patch fixes.

### DIFF
--- a/src/osgEarth/BuildGeometryFilter.cpp
+++ b/src/osgEarth/BuildGeometryFilter.cpp
@@ -1173,8 +1173,8 @@ BuildGeometryFilter::tileAndBuildPolygon(
 
     else
     {
-        // original tesselation approach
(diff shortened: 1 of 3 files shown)


emminizer:
Note, this is related to #2383 -- I was seeing the same thing on a recently updated Rocky 9 distro, and have a report from someone on a RHEL 8 distro. In both cases, the machines in my control were using the Mesa 23.1.4 driver, and the problem did not show up in earlier Mesa driver versions.

emminizer:
This is now OBE due to your alternate fix in https://github.com/gwaldron/osgearth/commit/e3ece4eed78f . Thanks for looking at this issue and helping us get a resolution so quickly!

PR #2402, opened 2023-12-11
Rex: Forcibly disable terrain morph for Mesa 23.1.4 driver

emminizer:
If using Mesa 23.1.4, disable terrain morphing. The 23.1.4 driver has been shown to have some sort of problem conveying the `vec4` coordinates for the `gl_MultiTexCoord0` through `oe_layer_tilec` value. This patch disables terrain morphing specifically only for the Mesa 23.1.4 driver.

If this affects future Mesa versions this can be expanded to cover more. in the future. This is intended as a precise change for the specific bug seen on two R8/R9 workstations with terrain morphing enabled. This is likely related to #2383 and is the terrain portion of the fix whereas #2401 is the image portion of the same fix.


emminizer:
This is now OBE due to your alternate fix in https://github.com/gwaldron/osgearth/commit/e3ece4eed78f . Thanks for looking at this issue and helping us get a resolution so quickly!

PR #2415, opened 2024-01-09
cesium-native: add a simple GLTF "baseColorFactor" support

remoe:
This adds a simple GLTF baseColorFactor support and fix a case when "baseColorTexture" is not initialized.

I understand that the GLTF osgEarth parser requires a new PBR shader to fully support all potential materials. However, this simple modification enables the display of colors in certain 3DTiles layers.


gwaldron:
Thanks Remo ;)

PR #2416, opened 2024-01-09
Draco: Fix CMake 3.23.2 configure error with unknown select_library_configurations macro

emminizer:
This fixes a failure on Linux on CMake 3.23.2. I'm not sure why this is not failing on other systems.


remoe:
@emminizer , thanks. This was from my last PR. I don't know why this worked without the includes on my configuration.

PR #2440, opened 2024-02-06
fix: gha macos build

remoe:
Update from MacOS 13 to 14


remoe:
It has a typo. I've updated the patch branch:

```
matrix:
        os: ['macos-14']
```

PR #2448, opened 2024-02-21
fix: weejobs macos build issues

remoe:
- std::mutex doesn't have a string as a constructor.
- initialize of a reference doesn't compile on clang

### DIFF
--- a/src/osgEarth/Containers
+++ b/src/osgEarth/Containers
@@ -632,7 +632,7 @@ namespace osgEarth { namespace Util
     struct PerObjectFastMap
     {
         PerObjectFastMap() { }
-        PerObjectFastMap(const std::string& name) : _mutex(name) { }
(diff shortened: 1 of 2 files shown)


gwaldron:
Hi Remo. Both of these occurred in clang because clang is evaluating templates before they are used.

Anyway, the first is a bug and I have fixed it in a separate push. Thanks.

The second however is not a problem with a default reference value, that is legal C++. The  problem is that the struct context type is being forward-declared (since it's evaluating the template before it's used) and you cannot assign a default value to an argument of a forward-declared type. So I will try to rearrange the code to prevent this.

remoe:
This is now solved by:

43237fd6c42b838d9e00ddcf46f19f8423fc081e
e00f167b0cfc755283f338b11ee5554f4bbc6c1f

PR #2458, opened 2024-03-02
Reset OGR order to standard;

MCA4213:
This commit allows:
1) Reset OGR geometries order to standard order.
2) Add geometryToIsoWKT function. (the geometryFromWKT is capable to read both wkt and isoWkt).

I have tested point, line, polygon geometries and are still loading correctly.
image

### DIFF
--- a/src/osgEarth/GeometryUtils
+++ b/src/osgEarth/GeometryUtils
@@ -31,6 +31,7 @@ namespace osgEarth
     namespace GeometryUtils
     {
         extern OSGEARTH_EXPORT std::string geometryToWKT( const Geometry* geometry );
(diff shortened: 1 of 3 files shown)


gwaldron:
Thanks for the PR!
Question: is there any reason not to default to ISO WKT? Should we just make `geometryToWKT` redirect to `geometryToIsoWKT`?

MCA4213:
I think it is possible to use only IsoWkt, because the old function is used
for compatibility.
Take a look on this:
https://gdal.org/api/vector_c_api.html

I am using spatialite and postgis and both of them use the iso format. (Pg
has an additional ewkt that is a little bit similar but with srid included:
not a standard i think).

Issue #2464, opened 2024-03-13
BUG - OSG version is not retrieved


AhmedAredah:
**Describe the bug**
Building the library on windows does not get the OSG version number.

**To Reproduce**
1. Build the library on Win 11 64 bit.
command:

`cmake -G "Visual Studio 16 2019" -A x64 .. -DGEOS_LIBRARY_DEBUG:FILEPATH="C:/Program Files/GEOS/lib/geos_cd.lib" -DOSGSHADOW_LIBRARY_DEBUG:FILEPATH="C:/Program Files/OSG/lib/osgShadowd.lib" -DOSGUTIL_LIBRARY
_DEBUG:FILEPATH="C:/Program Files/OSG/lib/osgUtild.lib" -DOSGSIM_LIBRARY_DEBUG:FILEPATH="C:/Program Files/OSG/lib/osgSimd.lib" -DOSGVIEWER_LIBRARY:FILEPATH="C:/Program Files/OSG/lib/osgViewer.lib" -DOSGSIM_LIBRA
RY:FILEPATH="C:/Program Files/OSG/lib/osgSim.lib" -DCMAKE_INSTALL_PREFIX:PATH="C:/Program Files/OSGEARTH" -DOSGTEXT_LIBRARY:FILEPATH="C:/Program Files/OSG/lib/osgText.lib" -DOSG_DIR:PATH="C:/Program Files/OSG" -
DOSGVIEWER_LIBRARY_DEBUG:FILEPATH="C:/Program Files/OSG/lib/osgViewerd.lib" -DOSGGA_LIBRARY:FILEPATH="C:/Program Files/OSG/lib/osgGA.lib" -DOSG_GEN_INCLUDE_DIR:PATH="C:/Program Files/OSG/include" -DOSGTEXT_LIBRA
RY_DEBUG:FILEPATH="C:/Program Files/OSG/lib/osgTextd.lib" -DOSGUTIL_LIBRARY:FILEPATH="C:/Program Files/OSG/lib/osgUtil.lib" -DOSGMANIPULATOR_LIBRARY_DEBUG:FILEPATH="C:/Program Files/OSG/lib/osgManipulatord.lib"
[...]

PR #2466, opened 2024-03-13
fix reading OSG version

AhmedAredah:


gwaldron:
Is this still relevant after our discussion in #2464 ?

AhmedAredah:
You can just discard it. Thank you. It worked without it after some time.

PR #2504, opened 2024-04-24
PBRMaterial: Fix Linux constructor build error

emminizer:
While Windows was happy to convert the `osg::Image*` to a `ref_ptr<osg::Image>` and construct a `Result` from that, gcc was unhappy. Explicitly constructing the `ref_ptr` fixes the issue.


gwaldron:
Thanks Dan.

PR #2506, opened 2024-04-25
Callbacks: Fixed remove() MSVC 2022 compilation issue. Incremented OSGEARTH_SOVERSION.

emminizer:
`osgEarth::Callback::remove()` now uses `std::vector::erase(iterator)` instead of trying to `erase()` the `pair` data element. MSVC issued compile errors when trying to call `remove()` before this.

`OSGEARTH_SOVERSION` incremented due to https://github.com/gwaldron/osgearth/commit/277332af from last week. Having the version update permits more code end user to compile cleanly with ifdefs in code.


gwaldron:
thanks Dan.

PR #2539, opened 2024-05-17
weejobs: Fix location of WEEJOBS_NO_DISCARD for gcc

emminizer:
Fixes compile error `error: standard attributes in middle of decl-specifiers` on gcc


gwaldron:
Thanks Dan :)

PR #2551, opened 2024-05-31
fix: macOS build

remoe:
MacOS build failed because of current CMake protobuf integration. And it had some minor compile issues.


remoe:
Here is the error for current implementation of Presets in clang without the fix of current PR:

`Default constructor of 'Presets' is implicitly deleted because field '_nodata' of const-qualified type 'const osgEarth::Config' would not be initialized
`

gwaldron:
Remo, what version of cmake do you use? I do not see any reference to `protobuf_MODULE_COMPATIBLE` in my installation (3.27.5)

gwaldron:
Remo, I pushed my own version of these fixes. Let me know. Thanks!

PR #2579, opened 2024-07-10
Updated to compile with Blend2D 0.11.4

kobalicek:
`BLContext::resetMatrix()` changed to `BLContext::resetTransform()`.


kobalicek:
Changes that match the following patch:

  - https://github.com/microsoft/vcpkg/pull/39730/files#diff-50490750cb741c861911cd5ffa598bfb50634b319c0445a190843107ac5070ad

gwaldron:
Appreciate this.
Any chance we can ifdef those to support both versions?

kobalicek:
It would be possible, but that previous version is basically out of support and has known bugs. I would be happier if projects used the latest.

PR #2593, opened 2024-07-18
Apps: Fixed ImGui for Windows and CollectTriangles apps

emminizer:
I was getting build errors in my Windows configuration without these fixes. Looks like they might have been overlooked during a recent refactor.


gwaldron:
Indeed they were...thanks Dan.

PR #2614, opened 2024-08-26
Capabilities: Remove unused Version include that does not compile

emminizer:
This was introduced in a recent refactor. Not sure how Windows is working here, as the "Version" file is in the build directory at `build_dir/build_include/osgEarth/Version`

Another fix is to update this back to `"osgEarth/Version"`. I suspect that removing this works simply because one of the other headers already includes it, but removing the header was a clean fix that compiled cleanly for me too.

Without a change here, Linux builds are continuing to fail. Again, unsure how Windows is apparently working.


gwaldron:
Thanks Dan. Yes. Local environment hid this problem.

Issue #2613, opened 2024-08-26
Add support for Cesium Ion self hosted


jasonbeverage:
We need to change some of the options in cesium native to support self hosted Cesium Ion so we can have per layer api tokens and server urls.

### Discussed in https://github.com/gwaldron/osgearth/discussions/2612

<div type='discussions-op-text'>

<sup>Originally posted by **remoe** August 26, 2024</sup>
Does support cesium-nodekit a self hosted Cesium ION with asset id's? It needs accessToken per layer and a server configuration entry for the cesium api server in the earth file?

In CesiumJS (javascript) one can add it as follow:

```js
const resource = await IonResource.fromAssetId(assetID, {
  accessToken: '<TOKEN>',
  server: 'https://example.ion'
});

const terrainProvider = CesiumTerrainProvider.fromUrl(resource);

```</div>

remoe:
@jasonbeverage I've done a first implementation. I've tested it with a single-user mode self-hosted ION mixed with default ION.

PR #2616, opened 2024-08-28
osgEarthCesium: support of Cesium Ion self-hosted

remoe:
This support Cesium Ion self-hosted. Tested with "single-user" mode. Use 'server' property in CesiumNative3DTiles to activate this. 

Example:
```xml
<CesiumNative3DTiles name="own dataset" >		
  <asset_id>1</asset_id>
  <server>https://api.example.ion/</server>
</CesiumNative3DTiles>
```

implementation for this: https://github.com/gwaldron/osgearth/issues/2613


jasonbeverage:
Thanks a bunch Remo!  This will definitely help customers that are running a self hosted Cesium Ion.

Issue #2642, opened 2024-10-15
custom shaderGenerator not being used by registry


gabr1gus:
osgEarth Version:  3.7.0 build 163

I was trying to use a custom shader generator so I would be able to create a different behavior for textures and shaderGen as the default is in conflict with the way that is used in my models, however, after creating this custom ShaderGenerator and using `osgEarth::Registry::instance()->setShaderGenerator()` to set this custom shadergen, I confirmed that no custom code was being called, even if using `osgEarth::Registry::shaderGenerator().run(model)` however, the custom code was being called if using the object directly, not through the registry.

For what I could find, this problem seem to be caused by osgEarth's use of ShaderGeneratorProxy.

All instances of registry that could return a instance of shaderGenerator, return instead an instance of shaderGenerator proxy, and in its constructor 
`ShaderGeneratorProxy(const ShaderGenerator* temp)
            : _instance( new ShaderGenerator(*temp, osg::CopyOp::SHALLOW_COPY) ) { }` 
It forces the creation of a new ShaderGenerator in specific, overruling any custom shadergenerator that may be passed to registry.

In my code I've "fixed" it by adding a function to shader generator:
`virtual ShaderGenerator* getCopy(const osg::CopyOp& copy) const { return new ShaderGenerator(*this, copy); } `
re-implementing this function in my custom ShaderGenerator as: 
`virtual ShaderGenerator* getCopy(const osg::CopyOp& copy) const { return new CustomShaderGenerator(*this, copy); } `
and altering shaderGeneratorProxy constructor to: 
`ShaderGeneratorProxy(const ShaderGenerator* temp)
            : _instance( temp->getCopy( osg::CopyOp::SHALLOW_COPY) ) { }`

gwaldron:
You are probably the first one is history to try this :) so I am not surprised there are issues.
Can you please submit a PR and we'll integrate it? Thanks.

PR #2644, opened 2024-10-16
Fix use of custom shaderGenerator in Registry.

gabr1gus:
Fixed use of ShaderGenerator in registry by altering ShaderGeneratorProxy constructor to initialize _instance wth new function `virtual ShaderGenerator::getCopy()` 

fix for issue #2642


gwaldron:
Thanks!

PR #2647, opened 2024-10-17
Update Cesium library list to match the current state of cesium-native

### LINKED ISSUE #2645: Cesium Native: Could NOT find csprng
osgEarth Version (required):
3.7.0

Description of the problem:
Could not find csprng while building osgearth with cesium-native support. 

What you have tried:
I followed the instructions in the documentation literally. Also tried to rebuild cesium-native and osgearth from the beginning.

Screenshot, code block, or data file that will help reproduce the issue:
image

ELeeScape:
The cesium-native project redid their dependencies a while back to use vcpkg and (I'm guessing) that caused osgEarth to be out of sync with what it needs.  This removes csprng which is no longer used, changes the names of modp_b64 -> modpbase64 and s2geometry -> s2, and adds a whole bunch of new libraries.  It also links in crypt32 on Windows, which is required by OpenSSL.  This should fix #2645.

I'm not entirely sure that every single one of the absl libraries is actually required, but after uncovering linker errors for three or four of them one at a time, I gave up and just dumped them all in there.

There might be a more elegant way to handle the need for linking crypt32 on Windows.  I see that cesium-native/install/lib/pkgconfig/libcrypto.pc contains the correct linker flags for each platform but I'm not very familiar with either cmake or vcpkg yet so I don't know how that's supposed to work.

After these fixes, the osgCesiumNative code builds, but it doesn't install correctly.  There are several dlls missing in the install/bin folder, such as osgViewer.dll, so none of the applications will run.  I'm not sure how to fix that yet.


ELeeScape:
Makes sense - I don't really know what I'm doing here so my PR was definitely brute-force.

PR #2665, opened 2024-12-03
WebP: sharpyuv only added if target exists, supporting older WebP

emminizer:
The added code works great for WebP 1.4, but WebP 1.2 (which we support on an older Linux distro) doesn't have sharpyuv.


jasonbeverage:
Thanks Dan!

PR #2683, opened 2025-01-08
Don't set prefix to lib64 on macOS

2xsaiko:
The lib dir on macOS is always called lib. Homebrew e.g. won't even link lib64.

(This should probably better use [GNUInstallDirs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html) anyway, but this is a quick fix the existing system)

### DIFF
--- a/cmake/oe_unix.cmake
+++ b/cmake/oe_unix.cmake
@@ -13,7 +13,9 @@ if(UNIX AND NOT ANDROID)
     find_package(Threads REQUIRED)
     
     # add 64 to the lib prefix.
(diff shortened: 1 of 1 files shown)


gwaldron:
Thanks -- it would be nice to consolidate any Apple-specific stuff in `cmake/oe_osx.cmake` or `cmake/oe_ios.cmake` -- do you mind seeing if this is feasible?

2xsaiko:
This is not Apple-specific at all, just a fix for macOS specifically because it is definitely broken there -- I wrote another approach.

PR #2685, opened 2025-01-10
MetaTile: Fixed erroneous typename usage

emminizer:
Linux build was failing due to extra `typename` values.

On MSVC sometimes this can be caught with enforcing conformance mode using `/permissive-` compile argument (added around MSVC 2017 15.5). I did not test if this particular case was caught by that flag on Windows, was only focused on the Linux build on gcc.


gwaldron:
Thanks Dan.

PR #2697, opened 2025-02-03
Linux: Fix reference compile errors

emminizer:
This returns an iterator in both cases, and is not able to be a reference. This fixes a compile error on gcc-13.


gwaldron:
thx.

PR #2703, opened 2025-02-07
LayersGUI: Fix missing include

emminizer:
CoverageLayer include was missing. Code was added recently to manipulate `CoverageLayer` in d08c83143 but it did not have an include. Build was failing on gcc-13 and MSVC 2022


gwaldron:
Oops, didn't actually mean to check this in yet, but what's done is done :)

PR #2746, opened 2025-04-23
Adopt non-deprecated GDAL 3 API; fix build with GDAL 3.11.0 debug

dg0yt:
Initially I only wanted to fix building the vcpkg port with GDAL 3.11.0 (https://github.com/microsoft/vcpkg/pull/45153): Debug builds of GDAL no longer export handle types as `void*`.

But after adding the first casts I stepped into complications and ended up porting away from API which is deprecated already in GDAL 2.0 (e.g. [OGR_DS_CreateLayer()](https://gdal.org/en/stable/doxygen/ogr__api_8h.html#a424d383a37fbeaea58acaea11717f320)).

I didn't do any runtime testing, only reading docs and building the vcpkg port.

For not blocking the gdal update in vcpkg, a quick review would be helpful. (The change is probably to invasive to be accepted as a patch.) I do understand that these change need proper review. The `static_cast` might hide problems (as does `void*`). Alternatively, it would be possible to replace the exposed `void*` with the actual handle types. Then the debug builds would unfold proper type checking.

### DIFF
--- a/src/osgEarth/OGRFeatureSource.cpp
+++ b/src/osgEarth/OGRFeatureSource.cpp
@@ -19,7 +19,7 @@ using namespace osgEarth;
 namespace osgEarth { namespace OGR
 {
     // helper function.
-    OGRLayerH openLayer(OGRDataSourceH ds, const std::string& layer)
+    OGRLayerH openLayer(GDALDatasetH ds, const std::string& layer)
     {
         OGRLayerH h = GDALDatasetGetLayerByName(ds, layer.c_str());
         if ( !h )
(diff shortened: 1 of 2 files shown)


gwaldron:
Thanks Kai. I did a quick review and what you did looks good. (Like you I didn't do any testing yet.)

Did you look at GDAL/GDAL.cpp and GDALDEM.cpp? I wonder if these need attention as well. The "GDAL" header has some questionable forward declarations of a couple handle types; we may need to rework that.

Exposing the real handle types in public headers would mean including GDAL itself in public headers, and this would require us to link GDAL as a public library which we would rather avoid.

gwaldron:
In the meantime, I tested this and it works, so I'll merge it now. Thanks!

PR #2753, opened 2025-04-28
GLUtils: Line stipple now unconditionally included for GL_LINE_STIPPLE

emminizer:
About ~L335 `GL_LINE_STIPPLE` is used. In a GL3 build of OSG, with `OSG_GL_FIXED_FUNCTION_AVAILABLE` not defined, the `GL_LINE_STIPPLE` value is also not defined. Adding `osg/LineStipple` include adds the define if necessary.

This ifxes a build error on Linux gcc 14.2 with OSG in GL3 configuration.

### DIFF
--- a/src/osgEarth/GLUtils.cpp
+++ b/src/osgEarth/GLUtils.cpp
@@ -16,11 +16,11 @@
 #include <osgViewer/GraphicsWindow>
 #include <osg/Texture2D>
 #include <osg/BindImageTexture>
+#include <osg/LineStipple>
 
 #ifdef OSG_GL_FIXED_FUNCTION_AVAILABLE
 #include <osg/LineWidth>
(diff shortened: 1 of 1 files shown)


remoe:
This doesn't compile with GL CORE enabled.

emminizer:
@remoe I am confused why the PR wouldn't compile with GL Core enabled. Note, I am compiling with GL Core and this is the reason for the patch.

The patch adds `osg/LineStipple` include (reference https://github.com/openscenegraph/OpenSceneGraph/blob/master/include/osg/LineStipple#L19 ), which explicitly defines `GL_LINE_STIPPLE` if it wasn't already defined. In GL Core mode, it wasn't defined, so the include adds it.

In original description I was using GL3 / GL Core profile interchangeably.

PR #2788, opened 2025-06-10
Fix pat in ModelLayer

leadcoder:
When both location and orientation is defined in the ModelLayer the orientation/pat is ignored because modelNodeParent now is set to the geo node, ie modelNodeParent should be pat if the model also should be rotated.


gwaldron:
Thank you!

PR #2789, opened 2025-06-10
Fix vsi support in gdal-driver

leadcoder:
The parsing for vsi-path is ignored, ie "source"-string is ignored.


gwaldron:
I feel like we could get rid of this if/else block altogether and just use `input = source;`
What do you think?

leadcoder:
Right, removing the if/else block would be the proper fix. i will update the PR

gwaldron:
Awesome, thx.

Issue #2752, opened 2025-04-28
can not run on macOS


fang-coco:
osgEarth Version (required):

➜  osgPlugins-3.7.0 osgearth_version --caps
[osgearth info] Hello, world.
[osgearth info] [Registry] Note: GDAL_DATA environment variable is not set
[osgearth info] [Capabilities] osgEarth Version:  3.7.3 build 173
[osgearth info] [Capabilities] GDAL Version:      3.10.3
[osgearth info] [Capabilities] OSG Version:       3.7.0
[osgearth info] [Capabilities] OSG GL3 Features:  yes
[osgearth info] [Capabilities] OSG FFP Available: no
[osgearth info] [Capabilities] CPU Cores:         8
[osgearth info] [Capabilities] GL_VENDOR:         Apple
[osgearth info] [Capabilities] GL_RENDERER:       Apple M3
[osgearth info] [Capabilities] GL_VERSION:        4.1 Metal - 89.3
[osgearth info] [Capabilities] GL CORE Profile:   yes
[osgearth info] Goodbye.

Description of the problem:

➜  tests git:(master) ✗ osgearth_imgui simple.earth
void StateSet::setGlobalDefaults() ShaderPipeline disabled.
void StateSet::setGlobalDefaults() ShaderPipeline disabled.
FRAGMENT glCompileShader "HexTiling.glsl_FRAGMENT" FAILED
FRAGMENT Shader "HexTiling.glsl_FRAGMENT" infolog:
ERROR: 0:233: '' : syntax error: incorrect preprocessor directive
ERROR: 0:233: '' : syntax error: unexpected tokens following #if preprocessor directive - expected a newline

glLinkProgram 0x6000001c6490"GDALImageLayer" FAILED
Program "GDALImageLayer" infolog:
ERROR: One or more attached shaders not successfully compiled

[osgearth warning] [VirtualProgram] Program will not link!
glLinkProgram 0x6000001d9a40"Terrain image layer" FAILED
Program "Terrain image layer" infolog:
ERROR: One or more attached shaders not successfully compiled

[osgearth warning] [VirtualProgram] Program will not link!
Warning: detected OpenGL error 'invalid operation' at after RenderBin::draw(..)
Warning: detected OpenGL error 'invalid operation' at after RenderBin::draw(..)
Warning: detected OpenGL error 'invalid operation' at after RenderBin::draw(..)
Warning: detected OpenGL error 'invalid operation' at after RenderBin::draw(..)
[...]

PR #2795, opened 2025-06-12
fix `HexTiling.glsl` compile error in MacOS

Lovely-XPP:
Relative to Issue #2752 (can not run on MacOS). The main problem is `FRAGMENT glCompileShader "HexTiling.glsl_FRAGMENT" FAILED`, and complete out is here
```bash
> osgearth_imgui simple.earth
void StateSet::setGlobalDefaults() ShaderPipeline disabled.
void StateSet::setGlobalDefaults() ShaderPipeline disabled.
FRAGMENT glCompileShader "HexTiling.glsl_FRAGMENT" FAILED
FRAGMENT Shader "HexTiling.glsl_FRAGMENT" infolog:
ERROR: 0:233: '' : syntax error: incorrect preprocessor directive
ERROR: 0:233: '' : syntax error: unexpected tokens following #if preprocessor directive - expected a newline
```

I found the problem is that the Macro `OE_ENABLE_HEX_TILER_ANISOTROPIC_FILTERING` is not defined. Therefore, I change the if condition from
```
#if OE_ENABLE_HEX_TILER_ANISOTROPIC_FILTERING
```
to
```
#if defined(OE_ENABLE_HEX_TILER_ANISOTROPIC_FILTERING) && (OE_ENABLE_HEX_TILER_ANISOTROPIC_FILTERING)
```
and the problem is solved.

### DIFF
--- a/src/osgEarth/HexTiling.glsl
+++ b/src/osgEarth/HexTiling.glsl
@@ -295,7 +295,7 @@ void ht_hex2colTex_optimized(
     vec2 st3 = st + ht_hash(vertex3);
 
 
-#if OE_ENABLE_HEX_TILER_ANISOTROPIC_FILTERING
+#if defined(OE_ENABLE_HEX_TILER_ANISOTROPIC_FILTERING) && (OE_ENABLE_HEX_TILER_ANISOTROPIC_FILTERING)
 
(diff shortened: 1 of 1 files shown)


gwaldron:
Thanks.

PR #2820, opened 2025-07-23
Add MinGW support for Windows compilation

adithyahk46:
### Summary

This PR adds support for building osgEarth using the MinGW toolchain on Windows. It includes several compatibility fixes and CMake enhancements to ensure successful compilation and runtime behavior under MinGW.

---

### Key Changes

- ✅ Replaced legacy `#ifdef WIN32` with standard-compliant `#ifdef _WIN32` across source files. Supports MSVC and MinGW.

- ✅ Added MinGW-specific memory alignment compatibility:

  ```cpp
  #ifdef _WIN32
      #include <malloc.h>
      #define memalign(x,y) _aligned_malloc(y, x)
      #define memfree(x) _aligned_free(x)
  #endif
  ```

- ✅ Implemented fallback for `_dupenv_s` for MinGW:

  ```cpp
  #ifdef __MINGW32__
  extern "C" errno_t _dupenv_s(...) { ... }
  #endif
  ```

---

- ✅ Updated CMake to:

  1. Link `dbghelp` when using MinGW:

     ```cmake
     if (MINGW AND WIN32)
         target_link_libraries(${LIB_NAME} PRIVATE dbghelp)
     endif()
     ```

  2. Link `ws2_32` on Windows using generator expression:

     ```cmake
     LIBRARIES
         $<$<PLATFORM_ID:Windows>:ws2_32>
     ```

  3. Set a `mingw_` prefix on plugin DLLs to match osgEarth loader expectations. osgEarth needs `mingw_` prefix, as it's built with `mingw_` prefix:

     ```cmake
     if (WIN32 AND MINGW)
         set_target_properties(${MY_TARGET} PROPERTIES PREFIX "mingw_")
     else()
         set_target_properties(${MY_TARGET} PROPERTIES PREFIX "")
     endif()
     ```

### DIFF
--- a/src/osgEarth/CMakeLists.txt
+++ b/src/osgEarth/CMakeLists.txt
@@ -913,7 +913,7 @@ if(UNIX)
 endif()
 
 # dbghelp?
-if(MINGW AND WIN32)
+if(WIN32)
     target_link_libraries(${LIB_NAME} PRIVATE dbghelp)
 endif()
 
--- a/src/osgEarth/Utils.cpp
+++ b/src/osgEarth/Utils.cpp
@@ -701,7 +701,6 @@ CustomRenderLeaf::render(osg::RenderInfo& renderInfo, osgUtil::RenderLeaf* previ
 #ifdef _WIN32
 #include <Windows.h>
 #include <dbghelp.h>
-#pragma comment(lib, "dbghelp.lib")
 #elif defined(__GNUC__)
 #include <execinfo.h>
 #include <cstdlib>


gwaldron:
Just a couple questions:

1. Why add _dupenv_s, and why put it in osgearth_imgui.cpp?
2. dbghelp.lib is added in Utils.cpp in a pragma comment but I'm guessing mingw doesn't pick that up. Perhaps just add your cmake change for all WIN32 builds (not just mingw) and we can remove the pragma?

adithyahk46:
1. ok Sorry, i found exact point that giving error, the problem is condition check in portable-file-dialogs.h , in the line 
static inline std::string getenv(std::string const &str)
{
#if _WIN32
    char *buf = nullptr;
    size_t size = 0;
    if (_dupenv_s(&buf, &size, str.c_str()) == 0 && buf)
    {
        std::string ret(buf);
        free(buf);
        return ret;
    }
    return "";
#else
    auto buf = std::getenv(str.c_str());
    return buf ? buf : "";
#endif
}

_dupenv_s only for MSVC builds. even for Mingw it falls in the first condition , just add the condition 
#if defined(_WIN32) && defined(_MSC_VER)


2. removed #pragma comment(lib, "dbghelp.lib") line from the Utills .cpp and in cmake change the condition from 
if (MINGW AND WIN32)  -> if(WIN32),

adithyahk46:
I think everything is ok now ? Or need changes

PR #2841, opened 2025-09-15
Add cstdint include to fix build errors on linux

mk401:
Linux builds on our end are failing on osgEarth main, simple include fix addresses the issue on usage of `std::uint64_t`


gwaldron:
Thanks!

PR #2842, opened 2025-09-16
Removed reference from auto to avoid binding a non-const ref to a tem…

mk401:
…poerary value returned by emplace()

Also picked up by our linux build machines (gcc 14)


gwaldron:
Thanks ;)

PR #2856, opened 2025-10-27
Blend2d: ifdef to support 0.20.0 version

vitaliikuzhdin:
Blend2D 0.20.0 introduced a new API that replaces camelCase methods with snake_case equivalents

Reference: https://github.com/blend2d/blend2d/commit/320a63aaf11c028a650a6b7d54cd3b7b12d8feed

It might be easier to update to the new API completely and include a compatibility header if an older version is detected. I also have not updated unreachable code.


gwaldron:
Down the road we can probably just support only the newer version but this works for now. Thanks!

PR #2859, opened 2025-11-10
Blend2d: ifdef to support 0.21.2 version

vitaliikuzhdin:
Blend2D 0.21.2 moved blend2d.h to blend2d/blend2d.h.

Reference: https://github.com/blend2d/blend2d/commit/def0d1238c3e5d0983bb848e5676049d829e435b
Related PR: https://github.com/gwaldron/osgearth/pull/2856

### DIFF
--- a/src/osgEarth/FeatureRasterizer.cpp
+++ b/src/osgEarth/FeatureRasterizer.cpp
@@ -17,7 +17,11 @@ using namespace osgEarth;
 #endif
 
 #ifdef USE_BLEND2D
+#if BL_VERSION >= 5378
+#include <blend2d/blend2d.h>
+#else
 #include <blend2d.h>
+#endif
 #endif
 
 #include <osgEarth/AGG.h>


gwaldron:
Thank you!

vitaliikuzhdin:
@emminizer, thanks for the review. I'll move the discussion here since the entire PR is about this tiny piece of code.

What you're saying makes perfect sense to me, but my patch still seems to work. I maintain both the [blend2d](https://aur.archlinux.org/packages/blend2d) and [osgearth](https://aur.archlinux.org/packages/osgearth) packages on AUR, and after submitting my first PR to fix the build with Blend2D `0.20.0`, I almost mindlessly submitted a similar one for `0.21.2` without realizing that `BL_VERSION` wouldn't be defined until the header is included. However, I tested the patch in a clean build, and I can still build osgEarth against the new Blend2D in a clean Arch chroot, which isn't the case without the patch. Well, that would be true if it didn't error out much later with something GDAL-related, but that's an unrelated issue I'll look into separately.

I'm not sure what's going on here, so I'll defer to your experience if you have the time. I'd appreciate any insight.

Edit: my patch does not, in fact, work. I was testing it without removing [the sed command](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=osgearth&id=9a0a0bf1d41cd2dd5b428cf07cd2afdf9c80ad04#n92) that I had previously added as a hotfix before writing the patch...

emminizer:
@vitaliikuzhdin thanks for taking a look at this. Note, I posted what I think is a good fix at https://github.com/gwaldron/osgearth/pull/2868

I'm confused where `BL_VERSION` might be coming from in your test build. I can say that in my standalone build with CMake on MSVC 2026, I don't appear to have `BL_VERSION` defined and don't see where it might come from except the `blend2d.h` header.

Just brainstorming, but the only way I can see that it might be defined is if some other header before this includes some third party library that includes the header (GDAL? Something else?). I can say with certainty that in my build configuration, none of my custom built third party libraries build against Blend2D.

I double checked to make sure it wasn't a version issue -- I definitely 100% do not have `BL_VERSION` defined where you're using it here, until after the header is included. Once the header is included, the ifdefs work. I think the cleanest, most compatible solution is simply to throw the test into CMake and let CMake decide whether to use the new include style or not. It's too bad there's not a compatibility header in the new version for Blend2D.

My vote still would be for https://github.com/gwaldron/osgearth/pull/2868 but I'm happy to update it if another path is desired.

gwaldron:
@vitaliikuzhdin My guess is that the old header might still be lurking somewhere in your include path and `BL_VERSION` is getting pulled in from there.

vitaliikuzhdin:
> Just brainstorming, but the only way I can see that it might be defined is if some other header before this includes some third party library that includes the header (GDAL? Something else?).

`blend2d` is an unofficial AUR package, no official Arch package would build against it, so that path seems unlikely. Building `osgearth` requires only two other AUR packages, [lerc](https://aur.archlinux.org/packages/lerc) and [python-sphinx-markdown-tables](https://aur.archlinux.org/packages/python-sphinx-markdown-tables), and neither has anything to do with Blend2D. That narrows the source of inclusion down to something coming directly from the osgEarth build itself.

> My guess is that the old header might still be lurking somewhere in your include path and BL_VERSION is getting pulled in from there.

My report comes from a clean chroot, it cannot have any leftovers by definition.
 
In any case, as much as it would be interesting to find the root cause, it is not that important. Thanks for correcting my approach and for investigating the issue.

PR #2876, opened 2025-12-08
Threading: shared_mutex requires mutex header

emminizer:
Fixes a Linux EL8 g++ 14.2 compile bug

### DIFF
--- a/src/osgEarth/Threading
+++ b/src/osgEarth/Threading
@@ -4,6 +4,7 @@
  */
 #pragma once
 #include <osgEarth/Common>
+#include <mutex>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


gwaldron:
Is this OBE with the shared_mutex fix from this morning?

emminizer:
I believe it is still an issue. I will double check and post back in 15-20 minutes.

emminizer:
It looks like I'm wrong, this is now OBE. Thanks for questioning so that I double checked.

gwaldron:
`Threading` includes `Common` which includes `PolyfillCPP14` which includes `<mutex>` and `<shared_mutex>`

Issue #2890, opened 2025-12-23
OverlayDecorator can inherit `DO_NOT_COMPUTE_NEAR_FAR` when computing the RTT parameters


surrealwaffle:
osgEarth Version (required): 3.7.2

### Description of the problem:

As part of [`OverlayDecorator::cullTerrainAndCalculateRTTParams()`](https://github.com/gwaldron/osgearth/blob/127c98ba3f73706b1cf753db46998d242d029ec1/src/osgEarth/OverlayDecorator.cpp#L478), the terrain subgraph is traversed by the cull visitor, with some intent to use that traversal to compute near-far values for a projection matrix for the RTT technique.

However, if `cv->getComputeNearFarMode()` is `DO_NOT_COMPUTE_NEAR_FAR`, then the computed near and far values are unchanged from `FLT_MAX` and `-FLT_MAX`, respectively. This results in complaints in `OpenSceneGraph` and some visual issues with certain techniques (e.g., lines projected using draping do not respect the specified width).

It is possible that `COMPUTE_NEAR_USING_PRIMITIVES` also results in similar issues (I have not tested this mode).

### What you have tried:

 1. Changing the camera to use `COMPUTE_NEAR_FAR_USING_BOUNDING_VOLUMES` works to resolve the issue in `OverlayDecorator`. However, this approach isn't viable for my current situation as my application has been tuned for some fixed near/far ratio and throws a bit of a visual fit otherwise.
 2. Patching `OverlayDecorator::cullTerrainAndCalculateRTTParams()` to save the old near-far compute mode, conditionally set it to `COMPUTE_NEAR_FAR_USING_BOUNDING_VOLUMES`, and restore it similar to `zSavedNear` and `zSavedFar`. This approach works reasonably well.

Is the approach described by (2) a viable resolution, or is the general expectation in osgEarth that near-far values should be recomputed as in (1)?

gwaldron:
I'd say that either (2) will work, or you could detect the DO_NOT_COMPUTE mode and plug in suitable N/F values directly.
Probably easier to just save/restore as in (2).

Note: the team will be out for the next couple weeks; if you want to submit a PR we can address it in the new year!

PR #2891, opened 2025-12-24
Make OverlayDecorator near/far compute mode aware

surrealwaffle:
Relevant issue: #2890

Currently, `osgEarth::OverlayDecorator::cullTerrainAndCalculateRTTParams()` is not sensitive to the current near/far compute mode and can result in default extremes being used for the near and far planes when the mode is `DO_NOT_COMPUTE_NEAR_FAR`.

This PR addresses this with the following changes:
 * save and restore the near/far compute mode across the cull traversal of the subgraph;
 * explicitly use `COMPUTE_NEAR_FAR_USING_BOUNDING_VOLUMES` to compute the near/far planes for the subgraph;
 * restore the saved near/far plane values without modification depending on the compute mode.


gwaldron:
thanks!

PR #2906, opened 2026-01-23
Fix GDAL API compatibility for GetMetadata return type

AhmedAredah:
Use auto to handle both older GDAL (char**) and newer GDAL (CSLConstList) return types from GetMetadata().


gwaldron:
Thanks.

AhmedAredah:
Anything to help. Thank you for maintaining it

PR #2934, opened 2026-03-11
Fix image layers rendering blank on macOS/Metal (sized internal texture format) within `ImageLayer.cpp`

Lovely-XPP:
Since v3.8.0, all image layers (OSM tiles, WMS, etc.) render as grey on macOS with Apple Silicon due to Metal/MoltenVK rejecting unsized GL internal formats passed to `glTexStorage2D`.

By the inspiration of copilot, another file `ImageLayer.cpp` is also needed to apply the patch, and test successfully in my M2 Macbook Pro.

# Changes
- `ImageLayer.cpp`
```cpp
// before: not set internalFormat 
auto mosaic = new osg::Image();
mosaic->allocateImage(cols, rows, layers, proto->getPixelFormat(), proto->getDataType());

// after: sized format applied so glTexStorage2D gets GL_RGB8, not GL_RGB
auto mosaic = new osg::Image();
auto pixelFormat = proto->getPixelFormat();
auto dataType = proto->getDataType();
auto internalFormat = proto->getInternalTextureFormat();

// Fix incorrect internal format if necessary: images loaded from disk often
// have a base (unsized) internal format like GL_RGB which is the same as the
// pixel format. We need a sized format like GL_RGB8 for use with glTexStorage2D
// (required on some platforms such as macOS/Metal via MoltenVK).
if (internalFormat == pixelFormat) {
    if (dataType == GL_UNSIGNED_BYTE)
    {
        if (pixelFormat == GL_RGB) internalFormat = GL_RGB8;
        else if (pixelFormat == GL_RGBA) internalFormat = GL_RGBA8;
        else if (pixelFormat == GL_RG) internalFormat = GL_RG8;
        else if (pixelFormat == GL_RED) internalFormat = GL_R8;
    }
    else if (dataType == GL_UNSIGNED_SHORT)
    {
        if (pixelFormat == GL_RGB) internalFormat = GL_RGB16;
        else if (pixelFormat == GL_RGBA) internalFormat = GL_RGBA16;
        else if (pixelFormat == GL_RG) internalFormat = GL_RG16;
        else if (pixelFormat == GL_RED) internalFormat = GL_R16;
    }
    else if (dataType == GL_FLOAT)
    {
        if (pixelFormat == GL_RGB) internalFormat = GL_RGB32F;
        else if (pixelFormat == GL_RGBA) internalFormat = GL_RGBA32F;
        else if (pixelFormat == GL_RG) internalFormat = GL_RG32F;
        else if (pixelFormat == GL_RED) internalFormat = GL_R32F;
    }
}
mosaic->allocateImage(cols, rows, layers, internalFormat, dataType);
```

### DIFF
--- a/src/osgEarth/TerrainTileModel.cpp
+++ b/src/osgEarth/TerrainTileModel.cpp
@@ -3,20 +3,8 @@
 * MIT License
 */
 
-#include <osg/GL>
 #include <osgEarth/TerrainTileModel>
 #include <osgUtil/IncrementalCompileOperation>
-
-// Fallback definitions for OpenGL 32-bit floating-point texture formats.
-// These are added to prevent "undeclared identifier" compilation errors 
-// on platforms with incomplete or outdated default GL headers (e.g., Windows and macOS).
-#ifndef GL_RGBA32F
-#define GL_RGBA32F 0x8814
-#endif
-
-#ifndef GL_RGB32F
-#define GL_RGB32F 0x8815
-#endif
 
 using namespace osgEarth;
 
--- a/src/osgEarth/TerrainTileModelFactory.cpp
+++ b/src/osgEarth/TerrainTileModelFactory.cpp
@@ -20,6 +20,17 @@
 #include <Superluminal/PerformanceAPI.h>
 #endif
 
+// Fallback definitions for OpenGL 32-bit floating-point texture formats.
+// These are added to prevent "undeclared identifier" compilation errors
+// on platforms with incomplete or outdated default GL headers (e.g., Windows and macOS).
+#ifndef GL_RGBA32F
+#define GL_RGBA32F 0x8814
+#endif
+
+#ifndef GL_RGB32F
+#define GL_RGB32F 0x8815
+#endif
+
 using namespace osgEarth;
 
 #define LABEL_IMAGERY "Terrain images"


jasonbeverage:
So just to be clear did the fix that copilot came up with not work for you?  It seems like your PR applies the same fix to the mosaicing path in ImageLayer::assembleImage but the mosaicing should run first and then the final texture should be ran through the code in TerrainTileModelFactory::createImageTexture that CoPilot came up with.

Lovely-XPP:
Yes, that's correct. I’ve verified this on my Mac by testing the commits one by one.

I identified that the issue started when the code in ImageLayer was changed from:

```cpp
mosaic->allocateImage(cols, rows, layers, GL_RGBA, GL_UNSIGNED_BYTE);
```
to
```cpp
mosaic->allocateImage(cols, rows, layers, proto->getPixelFormat(), proto->getDataType());
```
in commit [cf2887](https://github.com/pelicanmapping/osgearth/commit/cf2887a0a2f2131b900b6de3418d83d3f37708bf#diff-830ea30b213385d4865e3b7a5b972f4b2e021b1e24e4514e0141a584a69fc250R667#Line667)

It broke right after this change. Based on Copilot’s suggestion, ImageLayer requires a similar fix because the mosaicing path needs to ensure the pixel format and data type are handled consistently with how the final texture is created. Applying the fix only in TerrainTileModelFactory isn't enough if the source mosaic is already allocated incorrectly.

jasonbeverage:
This was resolved by 5d0fea1a5059602687ebeb7210b9f4557d5117e6

PR #2941, opened 2026-03-20
Update the enums for SilverLining cloud types.

fkane:
Sync up the enumeration for SilverLining cloud types with the current SilverLining API. Folks were getting the wrong clouds because of this.


gwaldron:
Thanks Frank.

PR #2998, opened 2026-08-26
LineDrawable: Fix stipple direction flip when segment crosses near clipping plane

emminizer:
Fixes a problem in SIMDIS where, when we zoom in close to an animated line, its stipple direction was randomly swapping the other direction. It no longer does that after this change. Tested against osgearth_drawables and everything looks the same and animates in the same direction.

Was seeing a case when zooming in that w was getting close to 0 as line approached near clipping plane, inverting the stipple and making "rotating" or "animated" lines appear to animate in reverse.

This fix manually clips the adjacent vertex against a near-plane approximation (w = 1e-4) to find a safe clip-space coordinate that remains in front of the camera, preserving the correct 2D pixel direction.


gwaldron:
Thanks Dan!

PR #3001, opened 2026-09-03
Fix a null access crash and buffer overflow issue found in WMS layer requests

RALaFrance222:
Encountered these potential crashes when working with WMS layers in SIMDIS. 

1. The ArcGISServer method was blindly accessing _imageLayer, and if it had been deleted was causing a crash to desktop
2. The HttpClient method was overflowing the char buffer if the input Request URL was longer than 256 characters, which seems quite easy to do

Verified both fixes functioned in osgearth_viewer and SIMDIS


gwaldron:
Thank you!

